PR #10, opened 2026-03-27
Updates to lammps set generators to use with new atomate2 flows

vir-k01:
This PR is part of an effort to include workflows for running LAMMPS via atomate2. The original PR was: https://github.com/materialsproject/pymatgen/pull/4368. A concurrent PR has been opened in atomate2 that has the flows (https://github.com/materialsproject/atomate2/pull/1185), while this PR aims to update the input set generators to match the inputs required there. These changes are based on initial work here: https://github.com/Matgenix/atomate2-lammps. Also tagging in @esoteric-ephemera who helped structure the code in this PR.

Major changes:
- Updated templates for running NVT/NPT/Minimization jobs with the upcoming flows in atomate2.
- Added in a data class to validate commonly utilized settings (based on the updated templates)
- The new input set generator is written (to the best of my ability) to provide input settings consistent with classical MD run through other codes on solids (matched defaults to ASE MD), and basic support for molecules.

Note: 
- I've introduced new classes for `BaseLammpsGenerator` and `LammpsForceField` which simplify the handoff of data to the atomate2 flows, while retaining the other generator classes in generators.py and the `ForceField` class in data.py; this was done to not break existing workflows and since these objects serve very different purposes/philosophies.

### DIFF
--- a/src/pymatgen/io/lammps/data.py
+++ b/src/pymatgen/io/lammps/data.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 import itertools
 import re
 import warnings
+from dataclasses import dataclass, field
 from io import StringIO
 from pathlib import Path
 from typing import TYPE_CHECKING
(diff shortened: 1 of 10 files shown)


shyuep:
Thanks. A quick question - are these changes backwards compatible?

vir-k01:
yup!

JaGeo:
@vir-k01 Linting still has some small problems, I believe.

@shyuep From atomate2 side, we would be very happy if this could be merged as soon as all tests pass.

shyuep:
Merged. The linintig issues are minor and I will fix from my end.

PR #13, opened 2026-03-31
Update outputs.py

hisvvhtek:
the original io/vasp/outputs.py line 5873 will lead to repeated computation of cder because the property cder is a method with @property. So direct computation will make it faster if we should go through many cders matrices with this function, just like my change


src/pymatgen/io/vasp/outputs.py
```diff
@@ -5870,7 +5870,7 @@ def get_orbital_derivative_between_states(
         Returns:
             a float value
         """
-        return self.cder[band_i, band_j, kpoint, spin, cart_dir]
+        return self.cder_real[band_i, band_j, kpoint, spin, cart_dir] + 1j * cder_imag[band_i, band_j, kpoint, spin, cart_dir]
```
shyuep: I think there is an error? Shouldn't it be self.cder_imag?

hisvvhtek:
yes, I made a mistake and you are right. I apologize for this mistake
________________________________
发件人: Shyue Ping Ong ***@***.***>
发送时间: 2026年4月15日 22:49
收件人: materialsproject/pymatgen-core ***@***.***>
抄送: hisvvhtek ***@***.***>; Author ***@***.***>
主题: Re: [materialsproject/pymatgen-core] Update outputs.py (PR #13)


@shyuep commented on this pull request.

________________________________

In src/pymatgen/io/vasp/outputs.py<https://github.com/materialsproject/pymatgen-core/pull/13#discussion_r3089844689>:

> @@ -5870,7 +5870,7 @@ def get_orbital_derivative_between_states(
         Returns:
             a float value
         """
-        return self.cder[band_i, band_j, kpoint, spin, cart_dir]
+        return self.cder_real[band_i, band_j, kpoint, spin, cart_dir] + 1j * cder_imag[band_i, band_j, kpoint, spin, cart_dir]


I think there is an error? Shouldn't it be self.cder_imag?

―

shyuep:
Thanks. Pls fix the error and push.

PR #25, opened 2026-04-15
Lazy import matplotlib.pyplot in structure_analyzer

lan496:
## Summary

Move the `matplotlib.pyplot` import from module scope into `VoronoiAnalyzer.plot_vor_analysis()`. structure_analyzer.py is mainly used for analysis, while matplotlib is only needed for one plotting method. Deferring the import reduces import overhead for users who do not use plotting. This would be consistent with https://github.com/materialsproject/pymatgen/pull/4128, which applied a similar lazy-import refactor elsewhere in pymatgen.

`matplotlib.pyplot` accounts for roughly 16-18% of the current import time of `pymatgen.core.structure_analyzer` in local profiling, though exact numbers vary by machine and system load.

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```

### DIFF
--- a/src/pymatgen/core/structure_analyzer.py
+++ b/src/pymatgen/core/structure_analyzer.py
@@ -8,7 +8,6 @@ import warnings
 from collections import defaultdict
 from typing import TYPE_CHECKING
 
-import matplotlib.pyplot as plt
 import numpy as np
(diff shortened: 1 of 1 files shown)


shyuep:
Thanks.

PR #28, opened 2026-04-16
updated RPA calculation parsing when ALGO=CHI

wuz75:
### Summary

This PR fixes several issues related to parsing VASP output from response function calculations (`ALGO=CHI`).

### Changes

- **`converged_electronic`**: Replace `ValueError` with a `warnings.warn()` when no ionic steps are found in an `ALGO=CHI` run. Response function calculations legitimately produce no SCF steps, so raising an exception was incorrect — a warning is more appropriate and allows parsing to continue.

- **`update_charge_from_potcar`**: Add `"CHI"` to the list of response-function `ALGO` tags that are excluded from charge updates via POTCAR. This is consistent with how other GW/BSE tags are handled.

### Motivation

`ALGO=CHI` runs were crashing during Vasprun parsing due to the hard `ValueError`, making it impossible to extract results (e.g. the dielectric tensor) from these calculations.

### DIFF
--- a/src/pymatgen/io/vasp/outputs.py
+++ b/src/pymatgen/io/vasp/outputs.py
@@ -670,7 +670,11 @@ class Vasprun(MSONable):
             real_partyz, real_partxz]], [[imag_partxx, imag_partyy, imag_partzz,
             imag_partxy, imag_partyz, imag_partxz]]).
         """
-        return self.dielectric_data["density"]
+        keys=self.dielectric_data.keys()
+        if "density" not in keys:
+            return self.dielectric_data['INVERSE MACROSCOPIC DIELECTRIC TENSOR (including local field effects in RPA (Hartree))']
+        else:
+            return self.dielectric_data["density"]
 
     @property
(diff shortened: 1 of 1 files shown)


shyuep:
Thanks for the contribution. But I am confused by there are 500+ modified files for a change in the RPA. Can you check what is happening and send a clean PR with only the necessary changes?

wuz75:
> Thanks for the contribution. But I am confused by there are 500+ modified files for a change in the RPA. Can you check what is happening and send a clean PR with only the necessary changes?

I ran this linter since I have it set up for other repos like emmet 
`pre-commit run --all-files`

is that alright? that's what the other 500 files are. The only rpa changes were made in `src/pymatgen/io/vasp/outputs.py‎`

shyuep:
Yeah, since it is impossible to evaluate the PR with 500 file changes, I would like to request that you make the changes in a separate PR that contains only the changes to the specific files. Thanks.

shyuep:
Thanks.

PR #29, opened 2026-04-18
Extending the plot_bands and plot_dos methods of BztPlotter for more control

deepanshuaggarwal51:
## Summary
Extended the `plot_bands` and `plot_dos` methods of `BztPlotter()` in `pymatgen.electronic_structure.boltztrap2` to give users more control over plotting and to make the two methods' return behavior consistent.

Changes:
`plot_bands`
- Added arguments matching those of `BztInterpolator.get_band_structure`, allowing the band structure to be configured directly from the plotting call.
- Now returns a `BSPlotter` instance instead of a plot object, mirroring `plot_dos`.

`plot_dos`
- Added a label argument to annotate the DOS in the plot with  "Total" as default.
- Added `**kwargs`, which are forwarded directly to `DosPlotter(**kwargs)` for full customization of the underlying plotter.

Why?
Previously, `plot_bands` and `plot_dos` had inconsistent return types and limited configurability. `plot_dos` already returned a `DosPlotter` instance, while `plot_bands` did not expose the underlying `BSPlotter`. The above changes:
- Put both methods on equal footing by returning their respective plotter classes, enabling further downstream customization by the user.
- Reduce boilerplate by letting users pass band-structure and DOS plotting options directly through the `BztPlotter` interface.

### DIFF
--- a/src/pymatgen/electronic_structure/boltztrap2.py
+++ b/src/pymatgen/electronic_structure/boltztrap2.py
@@ -1179,23 +1179,42 @@ class BztPlotter:
 
         return fig if ax is None else ax
 
-    def plot_bands(self):
+    def plot_bands(
+        self,
+        kpaths=None,
+        kpoints_lbls_dict=None,
+        density=20,
+    ) -> BSPlotter:
         """Plot a band structure on symmetry line using BSPlotter()."""
         if self.bzt_interp is None:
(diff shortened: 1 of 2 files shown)


shyuep:
Thanks!

deepanshuaggarwal51:
> Thanks!

Thank you for the review! Happy to contribute 🙌

PR #33, opened 2026-04-23
Split NumPy constraints at Python 3.13

hheei:
## Summary

Split the runtime NumPy dependency at Python 3.13.

- keep `numpy>=1.25.0,<3` for Python `<3.13`
- require `numpy>=2.1.0,<3` for Python `>=3.13`
- refresh `uv.lock` to match the new environment markers

Companion PR: materialsproject/pymatgen#4645

## Why

The VASP output tests were failing locally under a non-supported combination: Python 3.14 with NumPy 1.26.x.

This matches NumPy's support boundary:

- NumPy 1.26.x officially supports Python 3.9-3.12
- NumPy 2.1.0 adds Python 3.13 support
- newer NumPy 2.x releases cover later Python versions

This change aligns the dependency constraints with NumPy's official support window instead of allowing unsupported 1.x installs on newer Python versions.

## Validation

Ran the relevant VASP regression tests successfully on both Python 3.14 and Python 3.13:

- `tests/io/vasp/test_outputs.py::TestChgcar::test_init`
- `tests/io/vasp/test_outputs.py::TestAeccars::test_read_write_file`

Also ran:

- `uv lock --check`
- `uv run pre-commit run --files pyproject.toml uv.lock`


shyuep:
Thanks!

PR #47, opened 2026-05-13
Fix vec_to_surface

susumu-fujii:
## Summary

This PR fixes two issues in `GrainBoundaryGenerator.vec_to_surface`:

- select the reference component using the smallest absolute non-zero value instead of the numerically smallest value
- raise a `ValueError` when the vector cannot be reliably converted to Miller indices within tolerance

This avoids incorrect Miller index generation for vectors containing negative components and prevents silently returning unreliable indices from poor fractional approximations.

Regression tests are included for both cases.

## Todos

- none

## Checklist

* [x] Google format doc strings added. Check with `ruff`.
* [x] Type annotations included. Check with `mypy`.
* [x] Tests added for new features/fixes.
* [ ] If applicable, new classes/functions/modules have [`[duecredit](https://github.com/duecredit/duecredit)`](https://github.com/duecredit/duecredit) `@due.dcite` decorators to reference relevant papers by DOI

Additional notes:

- `uv run pytest tests/core/test_interface.py -k vec_to_surface` passes locally
- local `pre-commit` mypy reports an unrelated existing return type issue in `interface.py`


shyuep (commented):
> **Automated PR review generated by Claude (Opus 4.7)** on behalf of @shyuep. Based purely on the diff — no code was executed locally.

The fix is correct and addresses two real bugs in the old `vec_to_surface`:

1. **`np.argmin([i for i in vec if i != 0])` was buggy.** It returned an index into the *filtered* list and then indexed `index` (positions of nonzero components) with that — fine when shapes happen to align, but it also picked the smallest *signed* value, so `[-100, 3, 2]` would pick the `-100` component as the reference. The new `min(index, key=lambda i: abs(vec[i]))` picks the smallest absolute non-zero value, which is what you actually want as the denominator for Miller-index reduction. The added regression test `test_vec_to_surface_negative_components` documents this.
2. **Silent garbage on non-rational inputs.** `Fraction(...).limit_denominator(100)` will happily return a bad approximation for irrational ratios; raising `ValueError` when `|ratio - float(approx)| > 1e-8` is the right safety net. The matching test `test_vec_to_surface_invalid_fractional_approximation` covers it.

Minor things to consider, none blocking:

- **Tolerance.** `1e-8` against `limit_denominator(100)` is tight — it'll reject vectors that *could* reduce cleanly at a higher denominator. Worth a one-line comment that the 100 ceiling is a deliberate Miller-index sanity bound, not a numerical limitation. Also, expressing the tolerance relative to `|ratio|` (or `max(|ratio|, 1)`) would be more robust for very small components.
- **Return type annotation.** Signature says `tuple[int, ...]` but the function returns a `list` (via `miller`); the new test also asserts equality against a list. Pre-existing, but worth tightening either the annotation or the return.
- **CI hasn't run** on this PR (no status checks present). Please re-push or have a maintainer kick the workflow before merge.

Otherwise looks good — straightforward bug fix with tests for both branches.

susumu-fujii:
I updated the implementation to

- return a tuple consistently with the type annotation
- use a relative tolerance scaled by max(abs(ratio), 1)
- add a comment clarifying that the denominator limit is intended as a Miller-index sanity bound

The regression test was also updated accordingly.

shyuep:

- ✅ Both fixes are correct and minimal:
  - Picking the reference component by **smallest absolute non-zero value** (`min(index, key=lambda i: abs(vec[i]))`) instead of the numerically smallest avoids the previous mis-selection on vectors with negative components — the `[-100, 3, 2] → (-100, 3, 2)` test captures this exactly.
  - Raising on a poor fractional approximation (vs. silently returning a bogus Miller index) is the right call. The relative tolerance `tol = 1e-8 * max(abs(ratio), 1)` is sensible.
- ✅ The follow-up self-review (returning a tuple consistently, scaled relative tolerance, denominator-limit-as-sanity-bound comment) addresses what's worth tightening here.
- ✅ Type annotation widened to `tuple[float, float, float] | NDArray` matches existing call sites.
- ⚠️ `mergeable_state: behind` — branch needs a rebase on `main`. I'll attempt to update the branch automatically; if that fails due to fork permissions, please rebase locally.
- 💡 The two new regression tests are good, but consider one more: a vector with **two equal smallest-abs components** (e.g. `[1, 1, 0.5]`) to lock in deterministic `min(...)` tie-breaking behavior, since `min` returns the first index on ties and that's now load-bearing.
- 💡 The local `mypy` "return type issue in `interface.py`" you mention is pre-existing — agreed, not in scope for this PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

shyuep:
Thanks!

PR #48, opened 2026-05-13
Optimize volumetric data block parsing

hheei:
## Summary
- optimize generic VASP volumetric parsing in `VolumetricData.parse_file()`
- preserve `Locpot` and `Chgcar` parsing semantics, including spin-resolved volumetric blocks
- reduce `Locpot.from_file()` wall time on a large LOCPOT benchmark without adding a `Locpot`-specific code path

## Why
Large LOCPOT and CHGCAR files spend a substantial amount of time in generic volumetric text parsing. This change keeps the implementation generic while improving the large-file read path used by `Locpot.from_file()` and other `VolumetricData` consumers.

## Benchmark
Test environment:
- repo: `pymatgen-core`
- machine: local macOS M2
- file size: about 1.6 GB
- API measured: `Locpot.from_file()` only
- runs per implementation: 3

Parsed result:
- LOCPOT shape: `(300, 280, 576)`
- parsed keys: `['total', 'diff']`
- this LOCPOT is spin-resolved and contains 2 volumetric blocks, both of which are parsed by the current implementation

Comparison against the original implementation in this repository:

| Implementation | Run 1 (ms) | Run 2 (ms) | Run 3 (ms) | Mean (ms) |
|---|---:|---:|---:|---:|
| Original implementation | 23151.036 | 22663.119 | 22815.675 | 22876.610 |
| This PR | 8015.651 | 8104.559 | 8051.312 | 8057.174 |

Change relative to the original implementation:
- mean improvement: `14819.436 ms`
- relative improvement: `64.8%`

## Validation
- `uv run ruff check src/pymatgen/io/vasp/outputs.py`
- `uv run ruff check tests/io/vasp/test_outputs.py tests/performance/test_locpot_read.py`
- `uv run pyright src/pymatgen/io/vasp/outputs.py tests/io/vasp/test_outputs.py tests/performance/test_locpot_read.py`
- `uv run pytest -q tests/io/vasp/test_outputs.py -k Elfcar`

## Compatibility note
- `data_aug` now stores structured parsed arrays and metadata instead of the original raw text lines. This PR does not rely on the previous raw-line representation internally.


shyuep (commented):
> **Automated PR review generated by Claude (Opus 4.7)** on behalf of @shyuep. Based purely on the diff and current GitHub Actions state — no code was executed locally.

Nice perf win — switching from per-token Python parsing to a single `np.fromstring` over a precomputed block + `reshape(dim, order="F")` is the right shape and also dramatically simpler than the old `data_count // dim[0]` index unwinding. A few items worth a second look before merge:

1. **The block-size math hardcodes VASP's Fortran format and is now load-bearing.** `chars_per_val` (18/12), `vals_per_line` (5/10), and the `+3` trailing terminator come from `'(1X,E17.11)'` / `'(1X,G11.5)'` plus `WRITE(IU,*)' '`. If a file was written with a tweaked OUTCHG format, or by a non-VASP tool that mimics CHGCAR, the `np.fromstring(..., count=ngrid_pts)` will under-read and raise. The old parser tolerated any whitespace layout. Worth either (a) calling that contract out explicitly in the docstring, or (b) falling back to the old line-by-line path on size mismatch instead of raising.
2. **The new `llong` flag is a public API change.** `parse_file` gains a positional-style kwarg, and `Elfcar.from_file` now passes `llong=False` while `Locpot`/`Chgcar` pass `llong=True`. That's the correct VASP default split, but please confirm with a real ELFCAR — historically I've seen ELFCARs written with E17.11 too depending on build, and silently mis-parsing as `G11.5` would corrupt the grid without erroring.
3. **`diff` reconstruction changed.** Old: build `(3, N)` matrix, take norm, multiply by sign of dot with `[1.01, 1.02, 1.03]`. New: `np.sqrt(dx**2 + dy**2 + dz**2) * sign(1.01*dx + 1.02*dy + 1.03*dz)`. Mathematically equivalent, but it now keeps the 3D shape end-to-end and skips the reshape — good. No correctness concern, just flagging the silent behavior change for reviewers.
4. **CI:** `test (windows-latest, 3.13, lowest-direct, optional)` is failing and `test (macos-latest, 3.13, lowest-direct, optional)` is empty. They're marked optional, but please post the Windows failure log if it's actually touching this code path — `np.fromstring` deprecation behavior and locale-dependent float parsing have bitten this exact pattern before.
5. **No new tests.** A parser rewrite this aggressive should add at least one fixture for: (a) ELFCAR with `llong=False`, (b) a CHGCAR with the spin-polarized 4-dataset block, and (c) a malformed trailing block to confirm the new `ValueError` paths fire. Relying on the existing fixtures to catch regressions is thin.

Otherwise the refactor reads cleanly — happy to merge once Windows CI is understood and the `llong` default for ELFCAR is sanity-checked.

shyuep:

- ✅ Clear win: 55% wall-time improvement on a 1.6 GB LOCPOT is a meaningful speedup for a parser that's a real bottleneck on large CHGCAR/LOCPOT files. The approach (predict block byte-size from the VASP `FORM=` format + `np.fromstring`) is the right shape.
- ✅ Output shape is preserved by reshaping flat data with `order="F"`, matching VASP's x-fastest convention. The simplified spin-density `diff` reconstruction (`sqrt(dx²+dy²+dz²)*ref_sign`) is mathematically equivalent to and cleaner than the old `np.linalg.norm` + reshape dance.
- ⚠️ **Brittle assumption on character widths.** The `chars_per_val = 18 / 12` and `vals_per_line = 5 / 10` constants are tied to the exact VASP `WRITE` format (`(1X,E17.11)` / `(1X,G11.5)`). If VASP ever changes that format, or a third-party tool writes CHG-style files with a different field width, `block_chars` will be wrong and the `flat.size != ngrid_pts` guard fires. Worth a comment pointing at the VASP source location, and ideally a fallback path that re-parses tokenized if the fast path's `count` mismatches.
- ⚠️ **`Elfcar.from_file` passes `llong=False`** — is this verified against real ELFCAR output? VASP writes ELFCAR via the same `OUTCHG`-style routine but the `LLONG` flag for ELF is not obvious from the diff. A regression test with a known small ELFCAR (the existing test fixtures should have one) would lock this down.
- ⚠️ The new `llong` parameter is a public `staticmethod` parameter on `VolumetricData.parse_file`. Downstream subclasses that override `from_file` will need to thread this through. A short note in the changelog and a docstring example would help.
- 💡 The `vals_per_line * chars_per_val + 1` accounts for the trailing newline, and `remainder * chars_per_val + 3` for `WRITE(IU,*)' '` — the inline comment is good, but worth restating the +3 ("space + LF on POSIX, plus CR on Windows-newline LOCPOTs") since CRLF-encoded files could throw this off by 1 char per line. If `zopen` normalizes line endings in text mode, that's fine — confirm in a docstring.
- 💡 Patch coverage isn't on this PR yet (CI hasn't run), but the test plan only mentions `Locpot or Chgcar`. Please add at least one `Elfcar.from_file()` test and one negative test (truncated/short file → `ValueError`).

---
_Generated by [Claude Code](https://claude.ai/code)_

shyuep:
**🤖 Automated PR review (generated by Claude on behalf of @shyuep)** — re-review after "Add ELFCAR coverage and locpot benchmark".

- ✅ The rewrite is a clear improvement over the earlier `np.fromstring` approach: `_plain_loadtxt` derives the column count from the first data line instead of hardcoding VASP's `E17.11`/`G11.5` field widths — this resolves the brittleness flagged in the previous review. ELFCAR round-trip + truncated-file `ValueError` tests are exactly what was missing, and augmentation occupancies (including complex parts) are now genuinely parsed and round-tripped.
- 🛑 **Debug leftover committed.** `outputs.py` ends with `if __name__ == "__main__": Locpot.parse_file("/Users/supercgor/Downloads/pmg-benchmark/LOCPOT.vasp642.gz")` — a hardcoded personal path in a library module. Remove it.
- 🛑 **Stray prototype file.** `tests/performance/locpot_fromstring_prototype.py` is a standalone script (no test function) still using deprecated `np.fromstring` — an exploration leftover. Remove it or convert to a real test.
- ⚠️ **`data_aug` type changed — breaking.** It goes from `list[str]` of raw lines to `dict[int, NDArray]`. `write_file` is updated to match, but external consumers / serialization of `data_aug` will break. Flag this in the changelog.
- ⚠️ `arr = np.loadtxt(BytesIO(line), max_rows=1)  # No use?` — the `# No use?` comment means even the author is unsure. Either justify that branch or delete it.
- ⚠️ **Benchmark is stale.** The PR body's 64.8% figure was measured on the `np.fromstring` prototype; the committed code uses `np.loadtxt` (typically slower). Please re-benchmark the actual implementation.
- Minor: `remainer` → `remainder` typo in `_plain_loadtxt`. CI checks (tests + lint) are green. The branch is behind `main` — I could not auto-update it (fork permissions), please merge `main` in.

---
_Generated by [Claude Code](https://claude.ai/code) — automated review, no code executed locally._

---
_Generated by [Claude Code](https://claude.ai/code)_

shyuep:
**🤖 Automated PR review (generated by Claude on behalf of @shyuep)** — re-review after "Clean up volumetric parser follow-ups" / "Address remaining volumetric parser review threads".

- ✅ Prior blocking issues resolved: the hardcoded-path `__main__` debug block is gone, the `tests/performance/locpot_fromstring_prototype.py` exploration script is removed (net diff is now just `outputs.py` + `test_outputs.py`), the `remainder` typo is fixed, and legacy `list[str]` `data_aug` input is handled in `write_aug` with a dedicated regression test.
- ✅ Good additions: ELFCAR round-trip + truncated-file tests, CHGCAR/SOC round-trip assertions, and the `Vaspwave` 2-component spin-polarized path. The `diff` magnitude refactor is mathematically equivalent to the old `(3, N)` reshape and is cleaner.
- ⚠️ **Dead branch.** In `parse_file`, `elif b"." in line:` still computes `arr = np.loadtxt(BytesIO(line), max_rows=1)` and discards it — the new comment confirms nothing consumes it. Drop the parse and just `continue`/`pass`.
- ⚠️ **Scope creep.** `_parse_from_incar` and `get_band_structure` were switched from `pathlib` to `os.path`. The `_parse_from_incar` bare-filename fix is legitimate (and tested), but converting `get_band_structure` is unrelated to volumetric parsing — consider reverting it to keep the diff focused.
- ⚠️ For freshly parsed files `data_aug` still changes from `list[str]` to `dict[int, NDArray]`. `write_file` now accepts both, but please still note the parsed-output type change in the changelog.
- 💡 The PR description's speedup figure was measured on the old `np.fromstring` prototype — please re-confirm it reflects the committed `np.loadtxt` implementation.
- CI (lint + tests on all platforms) is green.

---
_Generated by [Claude Code](https://claude.ai/code) — automated review, no code executed locally._

shyuep:
Thank!

PR #61, opened 2026-05-18
Use ``math.lcm``/``math.gcd`` to avoid deprecation warnings and cleanup some redundant code

kavanase:
Updates usages of ``monty.fractions.gcd/lcm`` to avoid deprecation warnings with the latest ``monty``:
```
Category: FutureWarning
Message : gcd is deprecated, and will be removed on 2028-01-01
Use math.gcd instead.
Origin  : /opt/hostedtoolcache/Python/3.13.13/x64/lib/python3.13/site-packages/pymatgen/core/composition.py:1376
```

Also removes two cases of custom LCM functions which are redundant, to make code more compact.

### DIFF
--- a/src/pymatgen/command_line/enumlib_caller.py
+++ b/src/pymatgen/command_line/enumlib_caller.py
@@ -37,7 +37,6 @@ from typing import TYPE_CHECKING
 
 import numpy as np
 from monty.dev import requires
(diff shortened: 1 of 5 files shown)


shyuep:
Thanks.

PR #65, opened 2026-05-20
Speed up StructureMatcher.group_structures with Improved Hashing

ThomasWarford:
## Summary

- Update the `SiteOrderedIStructure.__hash__` function to consider fractional coordinates as well as composition. This decreases the number of hash collisions when `StructureMatcher._get_reduced_istructure` (which is decorated with `@lru_cache`)) is called repeatedly with structures of the same composition.
- `StructureMatcher.group_structures` calls `StructureMatcher._get_reduced_istructure` repeatedly.
- In some cases, this speeds up `SlabGenerator.get_slabs` significantly.
 
https://github.com/materialsproject/pymatgen-core/blob/9ef9e152b05a39643d25121c8743a0fffbd02dfd/src/pymatgen/core/surface.py#L1384

### Speedup
The outputs from [demo_hash_speedup.py](https://github.com/user-attachments/files/28058850/demo_hash_speedup.py):
group_structures on 80 structures (36-atom NaCl supercell)
  old hash (composition only): 4.274 s
  new hash (composition + coords): 0.259 s
  speedup: 16.5x

## Checklist

- [x] Tests added for new features/fixes: N/A?
- [x] If applicable, new classes/functions/modules have [`duecredit`](https://github.com/duecredit/duecredit) `@due.dcite` decorators to reference relevant papers by DOI ([example](https://github.com/materialsproject/pymatgen/blob/91dbe6ee9ed01d781a9388bf147648e20c6d58e0/pymatgen/core/lattice.py#L1168-L1172))

### DIFF
--- a/src/pymatgen/core/structure_matcher.py
+++ b/src/pymatgen/core/structure_matcher.py
@@ -86,12 +86,10 @@ class SiteOrderedIStructure(IStructure):
         each lookup requires fewer __eq__ calls.
         """
         sites_hash = hash(
-            tuple(
-                (site.species_string, *np.round(site.frac_coords*1000).tolist())
-                for site in self.sites
-            )
+            tuple((site.species_string, *np.round(site.frac_coords * 1000).tolist()) for site in self.sites)
         )
         return hash((super().__hash__(), sites_hash))
+
 
 class AbstractComparator(MSONable, abc.ABC):
     """
--- a/src/pymatgen/core/structure_matcher.py
+++ b/src/pymatgen/core/structure_matcher.py
@@ -77,9 +77,21 @@ class SiteOrderedIStructure(IStructure):
         return list(self.sites) == list(other.sites)
 
     def __hash__(self) -> int:
-        """Use the composition hash for now."""
-        return super().__hash__()
+        """Hash incorporating site positions to avoid LRU cache collisions.
 
+        The lru_cache on _get_reduced_istructure uses this hash as a key.
+        With a composition-only hash, structures sharing the same composition
+        all collide, requiring an O(N_sites) __eq__ call against each cached
+        entry to confirm a miss. Including rounded fractional coordinates means
+        each lookup requires fewer __eq__ calls.
+        """
+        sites_hash = hash(
+            tuple(
+                (site.species_string, *np.round(site.frac_coords*1000).tolist())
+                for site in self.sites
+            )
+        )
+        return hash((super().__hash__(), sites_hash))
 
 class AbstractComparator(MSONable, abc.ABC):
     """
--- a/src/pymatgen/core/structure_matcher.py
+++ b/src/pymatgen/core/structure_matcher.py
@@ -85,10 +85,9 @@ class SiteOrderedIStructure(IStructure):
         entry to confirm a miss. Including rounded fractional coordinates means
         each lookup requires fewer __eq__ calls.
         """
-        sites_hash = hash(
-            tuple((site.species_string, *np.round(site.frac_coords * 1000).tolist()) for site in self.sites)
-        )
-        return hash((super().__hash__(), sites_hash))
+        species = tuple(site.species_string for site in self.sites)
+        coords = np.round(self.frac_coords, 3).tobytes()
+        return hash((super().__hash__(), species, coords))
 
 
 class AbstractComparator(MSONable, abc.ABC):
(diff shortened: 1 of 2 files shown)


shyuep:

- ✅ Motivation is sound and the 16.5× speedup on `group_structures` is a real win — the `lru_cache` on `_get_reduced_istructure` was effectively running an O(N_sites) `__eq__` against every same-composition cached entry on each miss.
- ✅ Including `np.round(site.frac_coords * 1000)` gives ~1e-3 fractional-coord resolution, which comfortably exceeds the numerical noise typical structures carry and matches the spirit of `StructureMatcher`'s own tolerances. `__eq__`/`__hash__` remain consistent: equal structures still hash equally (same species + same rounded coords).
- 🔍 Nit (consider): the hash builds a Python tuple of per-site tuples — for very large structures this is O(N) and not negligible compared to `super().__hash__()`. If profiling shows it's a problem in other call sites, hashing a `.tobytes()` of the rounded coord array is significantly faster. Not blocking.
- 🔍 Nit: a tiny regression test pinning that two same-composition, different-coordinate structures get distinct hashes would lock in the optimization against future regressions in `__hash__`.
- ✅ CI green across the matrix.

LGTM on the substance — happy to merge once the (optional) regression test is added or as-is if you'd prefer to keep this minimal.

ThomasWarford:
Done :)

shyuep:
Sorry, I might need to revert this change. Based on the code, I don't think it actually matches what StructureMatcher is supposed to do. 

Say I have two structures that are identical, except that one structure has frac coords that are translated by an integer. This hash would essentially fail?

ThomasWarford:
If the second set of coords are translated by an integer they will have different hashes yes, so `StructureMatcher._get_reduced_istructure` will repeat the calculations.

But I believe the existing code would also repeat the calculation! The composition based hashes would match, but under the hood `lru_cache` will call `__eq__` against each structure with the same hash. `__eq__` only returns `True` if all sites are identical. Since all the site coordinates differ by an integer `__eq__` would return `False` and the calculation would be repeated. (one avenue for further speedup might be to update `__hash__` and `__eq__` to use relative coordinates `all_coords-all_coords[0]`)

At least this is how I think things work under the hood, I might be mistaken. `SiteOrderedIStructure` is only used as an input to `_get_reduced_istructure`, as far as I can tell.

shyuep:
I am not sure. But there is another sceniaro. Let's say I have two structures. One which is at [0.00049999999, 0, 0] and the oher at [0.00050000001,0,0]. Wouldn't np.round give them different hashes? 

```
>>> np.round(0.000500001, 3)
np.float64(0.001)
>>> np.round(0.0004999999, 3)
np.float64(0.0)
```

These are essentially the same structure to 3 decimal places. 

The rule of hashes is simple - the same object must have the same hash. Different objects can have either the same or different hashes. The more unique the hash, the better. But clearly, in the scenario above, we do not have the same hash for the same structure.

ThomasWarford:
Thanks for the detailed review.

On the rounding edge case: the hash/equality contract only requires that `a == b` implies `hash(a) == hash(b)`, not that "nearly equal" structures share a hash. `SiteOrderedIStructure.__eq__` checks `list(self.sites) == list(other.sites)`, which uses exact coordinate comparisons. So your two structures at `[0.00049999999, 0, 0]` and `[0.00050000001, 0, 0]` would also return `False` under `__eq__`. Different hashes for unequal objects is entirely valid, that's just a collision not occurring (fewer `__eq__` calls), which is the whole point.

If `__eq__` treats those two as the same object, then yes, the hash contract is violated,  but that isn't the case.

On the integer-translation case: agreed that two structures with coords differing by an integer would get different hashes and repeat the `_get_reduced_istructure` calculation. But as I mentioned above, they'd also fail `__eq__` and repeat the calculation under the old hash too - it just takes longer because every same-composition structure has to go through the full `__eq__` check before confirming a cache miss.

Since `SiteOrderedIStructure` is only used as an input to the `lru_cache`d function `_get_reduced_istructure`, the only correctness requirement is the hash/equality contract. A cache miss is always safe.

ThomasWarford:
In my mind hashing lets you check if two objects are identical, not semantically similar. For example two text files differing by a single space on a single line will get different sha256 hashes.

ThomasWarford:
TL:DR of above is that:
- `SiteOrderedIStructure` is only used as input to `_get_reduced_istructure`
- even if 0.0004999999 and 0.000500001 did get the same hash, since `0.0004999999!=0.000500001` the existing code would repeat the calculation after the equality check returns `False`.
- So, having different hashes for these two accelerates the code, by preventing equality checks.

shyuep:
OK. I will revert my revert. :-)

PR #67, opened 2026-05-21
Bugfix: do not assume projections in `HighSymmKPath.get_continuous_path`

esoteric-ephemera:
As in the title, as the `projections` are not a required field of `BandStructureSymmLine`, do not require them in `HighSymmKPath.get_continuous_path`


shyuep:
Thanks. Can we have a unittest to make sure the bug fix do not recur?

shyuep:

- ✅ The fix is correct. `get_continuous_path` previously did `bandstructure.projections[spin]` unconditionally, which raises `KeyError` when projections weren't computed (they're optional on `BandStructureSymmLine`). The `has_projections` guard (`isinstance(..., dict)` + all spins present) and the empty-dict fallback correctly handle the no-projections case.
- ✅ The signature/docstring type hints and the `BandstructureSymmLine` → `BandStructureSymmLine` typo fixes are good cleanup. `Spin` is already imported, so the annotation is fine.
- 🔍 Moving the projections loop inside the existing `for spin in spins` body is behavior-preserving.
- ⚠️ Still outstanding: @shyuep's request for a regression unittest hasn't been addressed (no test files in the diff; the last commit predates that comment). A test that calls `get_continuous_path` on a `BandStructureSymmLine` constructed *without* projections and asserts it returns successfully would pin this bug.
- Nit: `new_projections or {}` is redundant — on the no-projections path `new_projections` is already `{}`, so `projections=new_projections` suffices.

CI is green across all platforms. LGTM once a regression test is added.

shyuep:

Correct fix — `projections` is not a required field on `BandStructureSymmLine`, and the `has_projections` guard plus `projections=new_projections or {}` handles the empty case cleanly. Relocating projection accumulation into the per-spin bands loop is equivalent to the original. Type hints are a nice touch. CI is green. LGTM.

shyuep:

The `has_projections` guard is a reasonable fix for the no-projections case. However there appears to be a regression for spin-polarized inputs: the projections loop no longer has its own `for spin in spins:`, so it reuses the `spin` left over from the preceding bands loop and will only populate projections for the last spin (Spin.down), leaving Spin.up empty. Suggest wrapping it back in `for spin in spins:` inside the `if has_projections:` block and adding a spin-polarized + projections test. CI is green but likely doesn't cover that path.

shyuep:
🤖 _Automated PR review generated by Claude, posted on behalf of @shyuep. Based purely on static analysis of the diff and CI results — no code was executed locally._

Correct fix. Guarding projection handling with `has_projections` and returning `projections or {}` means `get_continuous_path` no longer assumes `bandstructure.projections` is a populated dict, which is the bug being fixed. Moving the projection concatenation under the `has_projections` branch is consistent.

- Minor: the `has_projections := isinstance(...) and all(...)` walrus is a little dense but fine.
- CI: the `lint` job is failing, but on `src/pymatgen/io/common.py` ("1 file would be reformatted") — a file this PR does not touch. This looks like unrelated formatting drift on the base branch, not an issue with your change. Rebase on latest main and/or run `ruff format` to clear it. All test matrices pass.
- Suggest a small regression test that builds a `BandStructureSymmLine` without projections and asserts `get_continuous_path` succeeds.

LGTM once lint is green.

PR #68, opened 2026-05-21
Refactor Structure/Molecule I/O dispatch to a plugin registry

shyuep:
## Summary

- Replaces the monolithic if/elif chains in `IStructure.{from_file,from_str,to}` and `IMolecule.{from_file,from_str,to}` with a plugin registry living in a new `pymatgen.io.registry` module. Each `pymatgen.io.<format>` now owns its own read/write glue and self-registers `StructureFormat`/`MoleculeFormat` handlers at module-import time.
- Built-in formats stay first-class: a static `format_name -> module_path` table lazy-imports the relevant io module on first use, so `from pymatgen.core import Structure` doesn't pay any extra import cost. External packages can plug in additional formats via the `pymatgen.io.structure_formats` and `pymatgen.io.molecule_formats` entry-point groups, or by importing a module that calls `register_structure_format(...)`.
- Existing user-facing behavior (format aliases, filename inference priority, error message wording, kwarg-warning semantics, default-to-JSON) is preserved. `IStructure._filter_kwargs` is kept as a thin shim around `pymatgen.io.registry.filter_kwargs`.
- `aims` and `fleur` (external namespace packages) get fallback shims in the registry so users with those packages installed keep working without a coordinated release.

## Plugin author API

```python
from pymatgen.io.registry import StructureFormat, register_structure_format, filter_kwargs

def _read(s, *, primitive=False, **kwargs):
    kwargs = filter_kwargs(SomeParser.from_str, kwargs)
    return SomeParser.from_str(s, **kwargs).structure

register_structure_format(StructureFormat(
    name="myfmt",
    patterns=("*.myf",),
    read_str=_read,
    write_str=lambda s, **kw: SomeWriter(s, **kw).get_str(),
    write_file=lambda s, fn, **kw: SomeWriter(s, **kw).write_file(fn),
))
```

External packages publish entry points in `pyproject.toml`:

```toml
[project.entry-points."pymatgen.io.structure_formats"]
myfmt = "my_pkg.io_module"
```

## Test plan

- [x] `uv run ruff check . && uv run ruff format --check .` — clean
- [x] `uv run mypy -p pymatgen` — 119 errors (was 120 on `main`; my change reduced by 1, no new errors in registry or refactored files)
- [x] `uv run pyright src` — 0 errors / 0 warnings
- [x] `uv run pytest tests` — **2725 passed**, 197 skipped (missing optional deps), 5 xfailed, 1 xpassed
- [x] New `tests/io/test_registry.py` (60 cases) covering registration, lookup by name and by filename, error paths, kwargs filtering, end-to-end roundtrips, and plugin-override-of-builtin
- [x] Manual smoke test: registering a custom `StructureFormat` and dispatching through the public `Structure.to/from_str/from_file` API succeeds and respects filename inference

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/pymatgen/core/structure.py
+++ b/src/pymatgen/core/structure.py
@@ -12,7 +12,6 @@ import functools
 import inspect
 import io
 import itertools
-import json
 import math
 import os
 import re
@@ -20,16 +19,13 @@ import sys
 import warnings
 from abc import ABC, abstractmethod
 from collections import defaultdict
-from fnmatch import fnmatch
 from typing import TYPE_CHECKING, Literal, cast, get_args, overload
 
 import numpy as np
 import orjson
 from monty.dev import deprecated
-from monty.io import zopen
 from monty.json import MSONable
 from numpy.linalg import norm
-from ruamel.yaml import YAML
 from tabulate import tabulate
 
 # scipy.cluster.hierarchy, scipy.linalg, and scipy.spatial are imported
@@ -2962,9 +2958,9 @@ class IStructure(SiteCollection, MSONable):
             fmt (str): Format to output to. Defaults to JSON unless filename
                 is provided. If specified, it overrides whatever the
                 filename is. Options include "cif", "poscar", "cssr", "json",
-                "xsf", "mcsqs", "prismatic", "yaml", "yml", "fleur-inpgen", "pwmat",
-                "aims".
-                Case insensitive.
+                "xsf", "mcsqs", "prismatic", "yaml", "yml", "fleur-inpgen",
+                "pwmat", "aims", or any format registered via
+                `pymatgen.io.registry`. Case-insensitive.
             **kwargs: Kwargs pass thru to relevant methods. This allows
                 the passing of parameters like `symprec` to the
                 `CifWriter.__init__ method` for generation of symmetric CIFs.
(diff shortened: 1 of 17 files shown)


shyuep:
Appreciate comments from @kavanase , @Andrew-S-Rosen and @esoteric-ephemera . I think adding more if-else for structure parsing is not going to work, especially when some of the lesser used io is implemented outside of pymatgen.

kavanase:
This looks like a good structured approach to me. 

Some minor things I noticed looking through:
- Some of the formatting in docstrings uses a single backtick when I think the intention is to use double backticks for in-line code format in Sphinx docs rendering (single backtick makes it italic); e.g. in `pymatgen.io.registry.py` header.
- Minor issue; redundant branch (`if filename` -> same error as otherwise) in https://github.com/materialsproject/pymatgen-core/blob/56d8276a6644d42d3ffb1499b24b7d356256c83e/src/pymatgen/io/registry.py#L417-L423
- The YAML IO functions accept `kwargs` but silently ignore them. I see this was actually also the case in the original code, but best to not do this (unless somehow required by the code structure?); https://github.com/materialsproject/pymatgen-core/blob/56d8276a6644d42d3ffb1499b24b7d356256c83e/src/pymatgen/io/registry.py#L559-L578
- Minor; https://github.com/materialsproject/pymatgen-core/blob/56d8276a6644d42d3ffb1499b24b7d356256c83e/src/pymatgen/io/registry.py#L209 is unreachable (and redundant) due to earlier: 
https://github.com/materialsproject/pymatgen-core/blob/56d8276a6644d42d3ffb1499b24b7d356256c83e/src/pymatgen/io/registry.py#L196
- Minor; there is a warning about `kwargs` being ignored for `fleur` when reading from str but not for the other `fleur` IO functions? I guess this should be thrown in each? https://github.com/materialsproject/pymatgen-core/blob/56d8276a6644d42d3ffb1499b24b7d356256c83e/src/pymatgen/io/registry.py#L669-L697

PR #69, opened 2026-05-23
Fix scientific notation list parsing/writing for `INCAR` files

kavanase:
From some [`ShakeNBreak`](https://shakenbreak.readthedocs.io/en/latest/) tests, I realised that the writing and parsing of `INCAR` flags with scientific notation in a list (e.g. [`ROPT`](https://vasp.at/wiki/ROPT); `1e-3 1e-3 1e-3`) failed, treating "e" as whitespace and converting it to `1 -3 1 -3 1 -3`. 

Essentially, list-type values are tokenized with a regex whose sub-pattern `-?\d+\.?\d*` does not support scientific-notation (treating "e" as whitespace). This PR extends the token regex to accept scientific notation: `-?\d+\.?\d*(?:[eE][-+]?\d+)?`.

Expanded some INCAR tests to catch this issue and test the fix.

### DIFF
--- a/src/pymatgen/io/vasp/inputs.py
+++ b/src/pymatgen/io/vasp/inputs.py
@@ -1042,11 +1042,14 @@ class Incar(UserDict, MSONable):
                     return int(str_)
 
                 output = []
-                tokens = re.findall(r"(-?\d+\.?\d*|[\.A-Z]+)\*?(-?\d+\.?\d*|[\.A-Z]+)?\*?(-?\d+\.?\d*|[\.A-Z]+)?", val)
+                # a single token is a number (incl. scientific notation, e.g. 1e-3) or an uppercase/dot
+                # string (e.g. .TRUE.); up to 3 tokens may be joined by "*" multipliers (e.g. 9*0.6)
+                num_or_str = r"-?\d+\.?\d*(?:[eE][-+]?\d+)?|[\.A-Z]+"
+                tokens = re.findall(rf"({num_or_str})\*?({num_or_str})?\*?({num_or_str})?", val)
                 for tok in tokens:
-                    if tok[2] and "3" in tok[0]:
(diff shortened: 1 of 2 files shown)


kavanase:
This issue arose with https://github.com/materialsproject/pymatgen/pull/4522; where `ROPT` was now treated as a list type instead of a string, hitting this regex which didn't support scientific notation. No tests present that caught this.

shyuep:

- ✅ **Regex fix is correct.** Extending the numeric token to `-?\d+\.?\d*(?:[eE][-+]?\d+)?` properly parses scientific-notation list values like `ROPT = 1e-3 1e-3`. The `num_or_str` factoring also makes the alternation easier to read.
- ✅ **Quiet bug fix on the nested multiplier.** The old guard `if tok[2] and "3" in tok[0]:` would only enter the `count1*count2*value` branch when the first token literally contained the character `"3"` — clearly a leftover/typo. `if tok[2]:` is the right condition (the inline comment now documents the intent).
- ✅ The new `2*1e-3` test confirms the multiplier syntax also handles scientific notation, and the `MAGMOM = 9*0.6` assertion implicitly exercises the existing multiplier path against the new regex.
- 🔍 Worth adding one more case to lock in the regex breadth: an uppercase exponent and an explicit positive sign, e.g. `ROPT = 1E-3 2.5E+2`, plus a negative-coefficient form (`-1e-4`). Cheap to add and prevents future regressions on the alternative forms.
- 🔍 Nit: the multi-line comment block above `tokens = re.findall(...)` is helpful — consider an analogous one-liner at the top of `smart_int_or_float_bool` describing the supported VASP value grammar, since this regex has now been a recurring source of subtle bugs (this PR + materialsproject/pymatgen#4522).

CI is green on all platforms. LGTM to merge as-is; the extra regex cases above are a nice-to-have, not blocking.

shyuep:
Thanks,

PR #72, opened 2026-05-25
fix: pbc not being preserved during structure transformations

deepanshuaggarwal51:
## Summary

PBCs (Periodic Boundary Conditions) set on a `Structure`'s `Lattice` is silently reset to `(True, True, True)` whenever certain structure transformations create a new `Lattice` object without forwarding the original `pbc`.  The most user-visible case is the `*`/`make_supercell` operator.

### Reproducing the bug
```python
from pymatgen.core import Lattice, Structure

# 2-D slab-like structure: non-periodic along z
a = 3
c = 20
latt = Lattice([[a, 0, 0], [0, a, 0], [0, 0, c]], pbc=(True, True, False))
s = Structure(latt, ["Si"], [[0, 0, 0]])

print(s.pbc)          # (True, True, False)  ← correct
print((s*(2,2,1)).pbc)   # (True, True, True)   ← BUG: pbc silently reset
```

### Comparison with ASE
[ASE (Atomic Simulation Environment)](https://ase-lib.org/) handles `pbc` correctly across **all** structure operations. Running the same workflow in both libraries makes the bug immediately concrete:

```python
from ase.atoms import Atoms

a = 3
c = 20
cell = [[a, 0, 0], [0, a, 0], [0, 0, c]]
pbc_flags = (True, True, False)           # slab: periodic in x,y only

ase_atoms  = Atoms("Si", positions=[[0, 0, 0]], cell=cell, pbc=pbc_flags)
print(ase_atoms.pbc)                      # (True, True, False)
print((ase_atoms * (2,2,1)).pbc)      # (True, True, False)
````

### Root Cause
The `IStructure.__mul__` creates a new `Lattice`  without passing `pbc=`, which defaults to `(True, True, True)`.

```python
# ❌ current (drops pbc)
new_lattice = Lattice(np.dot(scale_matrix, self.lattice.matrix))

# ✅ fix (forwards pbc)
new_lattice = Lattice(np.dot(scale_matrix, self.lattice.matrix), pbc=self.lattice.pbc)
```

### Diff
1. `IStructure.__mul__` and `make_supercell` via `*` operator
The primary bug. All supercell creation via `s * (2,2,1), s * 3`, or a full 3×3
scaling matrix drops `pbc`.
```diff
- new_lattice = Lattice(np.dot(scale_matrix, self.lattice.matrix))
+ new_lattice = Lattice(np.dot(scale_matrix, self.lattice.matrix), pbc=self.lattice.pbc)
```
`Structure.make_supercell` calls `__mul__` internally so it is fixed as a side-effect of this single change.

2. `Structure.apply_operation` — symmetry operations
Two branches (fractional-space and Cartesian-space) reconstruct `self._lattice`
after rotating it by a symmetry operation:

```diff
- self._lattice = Lattice(new_latt)
+ self._lattice = Lattice(new_latt, pbc=self._lattice.pbc)
```

```diff
- self._lattice = Lattice([symm_op.apply_rotation_only(row) for row in self._lattice.matrix])
+ self._lattice = Lattice([symm_op.apply_rotation_only(row) for row in self._lattice.matrix], pbc=self._lattice.pbc)
```

### Tests added (`tests/core/test_structure.py`)

- Appended to `TestIStructure.test_get_primitive_structure`

```python
def test_get_primitive_structure(self):
        ...

        # pbc must survive primitive-cell reduction
        coords = [[0, 0, 0], [0.75, 0.5, 0.75]]
        struct_pbc = IStructure(self.lattice_pbc, ["Si"] * 2, coords)
        struct_pbc_prim = struct_pbc.get_primitive_structure()
        assert struct_pbc_prim.pbc == (True, True, False)
```

- `TestStructure.setup_method` — appended `self.slab` to `setup_method`

```python
def setup_method(self):
    ...
    # 2-D slab: periodic in x,y only — used by pbc-preservation tests
    slab_lattice = Lattice([[3, 0, 0], [0, 3, 0], [0, 0, 20]], pbc=(True, True, False))
    self.slab = Structure(slab_lattice, ["Si"], [[0, 0, 0]])
```

- Appended to `TestStructure.test_apply_operation`

```python
def test_apply_operation(self):
    ...

    # pbc must survive symmetry operations
    op = SymmOp.from_axis_angle_and_translation([0, 0, 1], 90)
    slab = self.slab.copy()
    slab.apply_operation(op)    # Cartesian branch
    assert slab.pbc == (True, True, False)

    slab = self.slab.copy()
    slab.apply_operation(op, fractional=True)   # fractional branch
    assert slab.pbc == (True, True, False)
```

- Appended to `TestStructure.test_mul`

```python
def test_mul(self):
    ...

    # pbc must be preserved for every scaling form
    for scaling in [(2, 2, 1), 2, [[2, 0, 0], [0, 2, 0], [0, 0, 1]]]:
        assert (self.slab * scaling).pbc == (True, True, False)
```

- Appended to `TestStructure.test_make_supercell`

```python
def test_make_supercell(self):
    ...

    # pbc must survive in-place make_supercell
    slab = self.slab.copy()
    slab.make_supercell((2, 2, 1))
    assert slab.pbc == (True, True, False)
```

## Checklist

- [x] `IStructure.__mul__` now correctly forwards the pbc.
- [x] `make_supercell` now correctly forwards the pbc.
- [x] `apply_operation` - Symmetry operations on a structure now preserve the pbc.  
- [x] Unit tests for every changed method.
- [x] No change to the public API signature
- [x] Backwards-compatible for the overwhelmingly common `pbc=(True,True,True)` case.

### DIFF
--- a/src/pymatgen/core/structure.py
+++ b/src/pymatgen/core/structure.py
@@ -4316,7 +4316,7 @@ class Structure(IStructure, collections.abc.MutableSequence):
         """
         if fractional:
             new_latt = np.dot(symm_op.rotation_matrix, self._lattice.matrix)
-            self._lattice = Lattice(new_latt)
+            self._lattice = Lattice(new_latt, pbc=self._lattice.pbc)
 
             def operate_site(site):
                 return PeriodicSite(
@@ -4330,7 +4330,9 @@ class Structure(IStructure, collections.abc.MutableSequence):
                 )
 
         else:
-            self._lattice = Lattice([symm_op.apply_rotation_only(row) for row in self._lattice.matrix])
+            self._lattice = Lattice(
+                [symm_op.apply_rotation_only(row) for row in self._lattice.matrix],
+                pbc=self._lattice.pbc)
 
             def operate_site(site):
                 new_cart = symm_op.operate(site.coords)
--- a/src/pymatgen/core/structure.py
+++ b/src/pymatgen/core/structure.py
@@ -1158,7 +1158,7 @@ class IStructure(SiteCollection, MSONable):
         scale_matrix = np.array(scaling_matrix, int)
         if scale_matrix.shape != (3, 3):
             scale_matrix = scale_matrix * np.eye(3)  # (ruff-preview) noqa: PLR6104
-        new_lattice = Lattice(np.dot(scale_matrix, self.lattice.matrix))
+        new_lattice = Lattice(np.dot(scale_matrix, self.lattice.matrix), pbc=self.lattice.pbc)
 
         frac_lattice = lattice_points_in_supercell(scale_matrix)
         cart_lattice: NDArray = new_lattice.get_cartesian_coords(frac_lattice)
--- a/src/pymatgen/core/structure.py
+++ b/src/pymatgen/core/structure.py
@@ -4331,8 +4331,8 @@ class Structure(IStructure, collections.abc.MutableSequence):
 
         else:
             self._lattice = Lattice(
-                [symm_op.apply_rotation_only(row) for row in self._lattice.matrix],
-                pbc=self._lattice.pbc)
+                [symm_op.apply_rotation_only(row) for row in self._lattice.matrix], pbc=self._lattice.pbc
+            )
 
             def operate_site(site):
                 new_cart = symm_op.operate(site.coords)
--- a/tests/core/test_structure.py
+++ b/tests/core/test_structure.py
@@ -1046,6 +1046,10 @@ class TestStructure(MatSciTest):
         self.disordered = Structure.from_spacegroup("Im-3m", Lattice.cubic(3), [Composition("Fe0.5Mn0.5")], [[0, 0, 0]])
         self.labeled_structure = Structure(lattice, ["Si", "Si"], coords, labels=["Si1", "Si2"])
 
+        # 2-D slab: periodic in x,y only - used by pbc-preservation tests
+        slab_lattice = Lattice([[3, 0, 0], [0, 3, 0], [0, 0, 20]], pbc=(True, True, False))
+        self.slab = Structure(slab_lattice, ["Si"], [[0, 0, 0]])
+
     def test_calc_property(self):
         pytest.importorskip("matcalc")
         d = self.struct.calc_property("elasticity")
@@ -1463,6 +1467,16 @@ class TestStructure(MatSciTest):
         struct.apply_operation(op, fractional=False)
         assert struct.get_space_group_info() == spg_info
 
+        # pbc must survive symmetry operations
+        op = SymmOp.from_axis_angle_and_translation([0, 0, 1], 90)
+        slab = self.slab.copy()
+        slab.apply_operation(op)  # Cartesian branch
+        assert slab.pbc == (True, True, False)
+
+        slab = self.slab.copy()
+        slab.apply_operation(op, fractional=True)  # fractional branch
+        assert slab.pbc == (True, True, False)
+
     def test_apply_strain(self):
         struct = self.struct
         initial_coord = struct[1].coords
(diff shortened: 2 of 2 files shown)


shyuep:

The fix is correct and minimal — `Lattice.__init__` defaults `pbc=(True, True, True)`, so the supercell constructor was indeed silently overwriting the original PBC. Good reproducer and ASE comparison in the description, and all CI checks are passing.

Two small follow-ups worth considering before merge:

1. **Add a regression test.** A one-liner asserting `(s * (2, 2, 1)).pbc == (True, True, False)` for a slab-like cell would lock this in.
2. **Audit other call sites.** The PR title says "structure transformations" (plural), but only `IStructure.__mul__` is patched. A quick grep for `Lattice(` inside `pymatgen/core` (e.g. in `apply_strain`, `apply_operation`, `make_supercell`-adjacent paths, `Lattice.scale`, deformation tensors) is worth doing to confirm no other constructor drops `pbc`. If they do, please fix them in the same PR — otherwise users hit the same bug via a slightly different route.

Otherwise LGTM once a test is added.

shyuep:
Thanks for fixing. Can you add a unittest to ensure the roundtrip is properly done and to ensure that people working on the code in future do not inadvertently make a change that causes this to recur. Thanks.

deepanshuaggarwal51:
> Thanks for fixing. Can you add a unittest to ensure the roundtrip is properly done and to ensure that people working on the code in future do not inadvertently make a change that causes this to recur. Thanks.

Thank you for the review. Good call on adding a regression test. I have just pushed an update with appropriate checks that validate the full round-trip flow. This should safely protect the fix against any inadvertent changes down the line. Let me know if the implementation looks good to you!

Additionally, the methods where I am not sure about forwarding the PBC:
1. `Structure.apply_strain` - A small amount of strain will naturally preserve the PBC, but there may be cases where a user doesn't want the original structure's PBC carried over. Let me know if propagating the PBC (i.e., carrying it through to the strained structure) and adding a PBC lock-in unit test to cover this behavior makes sense here.

3. `IStructure.interpolate` 

```python
def interpolate(self, ...) -> list[Self]:
...
if not (interpolate_lattices or self.lattice == end_structure.lattice):
            raise ValueError("Structures with different lattices!")
...
```
If `interpolate_lattices=False`, it specifically checks for lattice equivalence and hence the PBCs associated with the initial and final structures. If the PBCs do not match, it raises a `ValueError`; if they do, it generates images with a silent reset of the PBCs to [True, True, True] of in-between lattices.

On the other hand, if `interpolate_lattices=True`, it completely skips the PBC check and generates interpolated images even if the initial structure is 3d periodic and the final structure is a slab. For instance, 

```python
coords = [[0, 0, 0], [0.75, 0.5, 0.75]]
struct = Structure(Lattice(
            [
                [3.8401979337, 0, 0],
                [1.9200989668, 3.3257101909, 0],
                [0, -2.2171384943, 3.1355090603],
            ]
        ), ["Si"] * 2, coords)
coords2 = [[0, 0, 0], [0.5, 0.5, 0.5]]
l2 = Lattice.from_parameters(3, 4, 4, 100, 100, 70)
l2.pbc = [True, True, False]
struct2 = Structure(l2, ["Si"] * 2, coords2)
int_s = struct.interpolate(struct2, 2, interpolate_lattices=True)
```

Let me know if patching the PBC forward inside the for loop, where the new lattice is created, is fine here as well.

shyuep:
**🤖 Automated PR review (generated by Claude on behalf of @shyuep)** — re-review after the 2026-05-26 commits.

- ✅ Latest commits round out the fix nicely: `Structure.apply_operation` now forwards `pbc` in both the fractional and Cartesian branches, and explicit pbc-preservation tests have been added for `get_primitive_structure`, `apply_operation`, `__mul__`, and `make_supercell`. CI is green across all matrix runs.
- ✅ The implementation pattern is the minimal correct one — every `Lattice(...)` reconstruction inside the structure now threads `pbc=self.lattice.pbc` through. ASE-parity argument in the PR description holds.
- 🟡 Minor: the new test `slab` fixture now lives in `setup_method` and is rebuilt for every test in `TestStructure`, even tests that don't touch pbc. It's cheap (single Si site), so fine, but a class-level fixture (or just inlining in the three pbc tests) would be slightly cleaner. Not a blocker.
- 🟡 Worth a quick scan for *other* call sites that reconstruct a `Lattice` without forwarding `pbc` — e.g. anywhere `Lattice(self.lattice.matrix.dot(...))` or similar appears in `structure.py` / transformations. The three sites fixed here are the user-visible ones, but if any internal helper builds a fresh `Lattice` from `self.matrix` alone, the same bug class can resurface. Not required for this PR, but a follow-up grep would be cheap insurance.

Approve as-is — this is a clean bugfix with good test coverage and matches ASE semantics.

shyuep:
Thanks for the fix!

deepanshuaggarwal51:
> Thanks for the fix!

Thank you for merging.

PR #73, opened 2026-05-27
Fix bug in get_projected_plots_dots

mias0356:
## Summary

Major changes:

- feature 1: `get_projected_plots_dots` does not give a single plt.Axes object for `ax` when one element and multiple orbitals are inputted for `dictio` and single-column subplots are expected. 
image
image

- fix 1: Fixed typo in line 1021 where `n_cols == 1` is repeated twice 
`if n_cols == 1 and n_cols ==1` changed to `if n_cols == 1 and n_rows ==1`. 
`elif ~` changed to include the cases where n_cols == 1 or n_rows == 1

## Todos

If this is work in progress, what else needs to be done? N/A

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```


shyuep:
Thanks.

PR #74, opened 2026-06-02
Fix bug in _get_projections_by_branches_patom_pmorb

mias0356:
## Summary

Major changes:

- feature 1: `get_projected_plots_dots_patom_pmorb` plots do not reflect the correct projections of `Spin.down` (cyan). 
For example, spin.down projection of Cr in band 835 (or 834 by index) is much smaller than spin.up in this band structure:
image  
but their marker sizes, which should reflect the difference in projections, look almost the same in the projected plot:
<br>
image

- fix 1: `Spin.up` changed to `Spin.down` in line 1372 ( `_get_projections_by_branches_patom_pmorb`)
image

## Todos

If this is work in progress, what else needs to be done? N/A  

## Checklist

```sh
pip install -U pre-commit
pre-commit install
```


shyuep:
Thanks!

PR #24, opened 2026-04-14
Fix Wrong Oriented Cell When Calling SlabGenerator.get_slab

ThomasWarford:
## Summary
Copy of https://github.com/materialsproject/pymatgen/pull/4598

- Only take primitive slab cell if corresponding oriented unit cell is found.
- Add `reduce=True` keyword argument to `Structure.get_primitive_cell`. If `false`, `Structure.get_reduced_structure()` is not applied during recursive search for primitive cell.
- When searching for primitive cell of a slab and its corresponding orientated unit cell (OUC), use `reduce=False`.
    - For MgO(111) example in #4597 this allows the oriented unit cell to find primitive cell candidates which meet the constraints, otherwise it can't. 
    - Someone who's more knowledgable about Niggli reduction might be able to rationalise this.

## Checklist

- [x] Google format doc strings added. Check with `ruff`.
- [x] Type annotations included. Check with `mypy`.
- [x] Tests added for new features/fixes.
- [x] If applicable, new classes/functions/modules have [`duecredit`](https://github.com/duecredit/duecredit) `@due.dcite` decorators to reference relevant papers by DOI ([example](https://github.com/materialsproject/pymatgen/blob/91dbe6ee9ed01d781a9388bf147648e20c6d58e0/pymatgen/core/lattice.py#L1168-L1172))

Closes #4597

PR #78, opened 2026-06-10
pass reduce kwarg to recursive IStructure.get_primitive_structure calls

ThomasWarford:
## Summary

Fix a bug introduced in #24 (sorry!) - when `IStructure.get_primitiive_structure` was called with `reduce=False` reduce was not passed to recursive `get_primitive_structure` calls.

This fix means SlabGenerator finds smaller slabs when `primitive=True`.

## Checklist

- [x] If applicable, new classes/functions/modules have [`duecredit`](https://github.com/duecredit/duecredit) `@due.dcite` decorators to reference relevant papers by DOI ([example](https://github.com/materialsproject/pymatgen/blob/91dbe6ee9ed01d781a9388bf147648e20c6d58e0/pymatgen/core/lattice.py#L1168-L1172))

### DIFF
--- a/src/pymatgen/core/structure.py
+++ b/src/pymatgen/core/structure.py
@@ -2776,6 +2776,7 @@ class IStructure(SiteCollection, MSONable):
                         tolerance=tolerance,
                         use_site_props=use_site_props,
                         constrain_latt=constrain_latt,
+                        reduce=reduce,
                     )
                     if reduce:
                         primitive = primitive.get_reduced_structure()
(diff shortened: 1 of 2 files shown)


shyuep:
The tests are failing due to some tolerance factor with adsorbate finder. There is no such failure on main. So can you check if there is a need to update the test tolerances? Thanks.

ThomasWarford:
The latest commit should fix this by specifying `atol`

PR #82, opened 2026-06-26
improve ewald vector ops

Bud-Macaulay:
: ]

Further vectorisation of the expensive ops `_calc_recip` and `_calc_real_and_point`.

As often with these vector ops there is some reduction in readability but in my testing of systems with 100+ atoms there was a +10x speedup.

Additionally I made some small changes to use np.sum when accessing some methods where appropriate. This is more performant than sum(sum(...)), but is likely a very small win. 

tests passed locally - there is no benchmark test file that i could see.

Happy to discuss

### DIFF
--- a/src/pymatgen/core/ewald.py
+++ b/src/pymatgen/core/ewald.py
@@ -135,7 +135,7 @@ class EwaldSummation(MSONable):
         for idx in removed_indices:
             total_energy_matrix[idx, :] = 0
             total_energy_matrix[:, idx] = 0
-        return sum(sum(total_energy_matrix))
+        return total_energy_matrix.sum()
 
     def compute_sub_structure(self, sub_structure, tol: float = 1e-3):
         """Get total Ewald energy for an sub structure in the same
@@ -177,7 +177,7 @@ class EwaldSummation(MSONable):
             output.extend(f"unmatched = {site}" for site in sub_structure if site not in matches)
             raise ValueError("\n".join(output))
 
-        return sum(sum(total_energy_matrix))
+        return total_energy_matrix.sum()
 
     @property
     def reciprocal_space_energy(self):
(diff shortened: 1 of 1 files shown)


shyuep:
🤖 **Automated PR review** (generated by Claude on behalf of @shyuep). Based purely on the GitHub diff — no code was executed.

Checked the algebra: the rewrite is equivalent to the original.

- Reciprocal space: `U=(cos+sin)·w`, `V=(cos−sin)·w`, `e_recip=(Uᵀ·sin + Vᵀ·cos)/√2` correctly reproduces `Σ_G w·sin(G·r_j + π/4 − G·r_i)` via the angle-addition identity, replacing the per-G Python loop with two matmuls. The force expression `prefactor·q_i·Σ_G 2w(s_real·sin − s_imag·cos)·G` matches the original term-by-term.
- Real space: `np.add.at(e_real[:, idx], js, contrib)` is a correct (and faster) replacement for the inner `for key in range(n_sites)` loop — the column slice is a true view and duplicate images accumulate. `np.fill_diagonal(...)` and the `.sum()` cleanups are fine.

Main ask before merge: **this is numerically sensitive code and the PR changes no tests.** Please add a regression test asserting energies and forces match reference values (including a charged cell and `compute_forces=True`) within tolerance.

Note: CI has not run on this branch yet — recommend confirming the existing Ewald tests pass before merging.

Bud-Macaulay:
The test file tests/core/test_ewald.py sufficiently tests for regression on complex cases and indicates no regression therefore i think this should be good to merge.

shyuep:

Good vectorization — matmul reformulation of the reciprocal sum plus `.sum()`/`np.fill_diagonal` cleanups. Because this rewrites the core Ewald math, it is merge-blocking on validation:
- Please add/confirm a regression test comparing `e_recip`, `forces`, and `total_energy` against the previous loop implementation on a known structure. The `sin(G·r_j + π/4 − G·r_i)` → two-matmul identity and the trailing `*2**0.5` scaling are easy to get subtly wrong.
- **No CI checks present** — existing `ewald` tests must pass before merge.

Logic looks plausible, but numerical equivalence cannot be confirmed from static review alone.

shyuep:
Thanks!

Issue #83, opened 2026-06-26
Consider switching to KDTree in MoleculeStructureCompariator module


Bud-Macaulay:
### Feature Requested

Currently, the MoleculeStructureComparitor calls the private method `_get_bonds(...)` which initialises a i,j pairs matrix across the whole input structure.

This results in large molecules hanging while this pair_dists calculation is performed (the majority of the comparisons will be culled in the proceeding forloop)

I suggest instead using scipy.spatial KDTree (a module that is used in Structure already) to change the algorithm from O(n^2) to O(nlog(n)).
```
import time
import numpy as np
from pymatgen.core import Molecule

from pymatgen.core.molecule_structure_comparator import MoleculeStructureComparator

n = 5000

coords = np.zeros((n, 3))
coords[:, 0] = np.arange(n) * 1.54

mol = Molecule(["C"] * n, coords)

comp = MoleculeStructureComparator()

t0 = time.perf_counter()
bonds = comp._get_bonds(mol)
t1 = time.perf_counter()

print(f"{len(mol)=}")
print(f"{len(bonds)=}")
print(f"time = {t1 - t0:.3f} s")

print(bonds[0:10])
```
Outputs:
```
len(mol)=5000
len(bonds)=4999
time = 54.739 s
[(0, 1), (1, 2), (2, 3), (3, 4), (4, 5), (5, 6), (6, 7), (7, 8), (8, 9), (9, 10)]
```

### Proposed Solution

Switch to pre-eliminate far apart pairs using KDTree.
```
    def _get_bonds(self, mol, max_cutoff=20):
        """Find all bonds in a molecule.

        Args:
            mol: pymatgen Molecule object.
            max_cutoff: The maximum radial distance and Angstrom to search for bonds

        Returns:
            List[tuple[int, int]]: bonded atom pairs.
        """
        n_atoms = len(mol)
        if self.ignore_ionic_bond:
            covalent_atoms = [idx for idx, sp in enumerate(mol.species) if sp.symbol not in self.ionic_element_list]
        else:
            covalent_atoms = list(range(n_atoms))

        unavailable_elements = set(mol.composition.as_dict()) - set(self.covalent_radius)
        if unavailable_elements:
            raise ValueError(f"The covalent radius for element {unavailable_elements} is not available")

        coords = mol.cart_coords[covalent_atoms]

        # Use a KDTree to avoid N^2 operations in building the pair list
        tree = cKDTree(coords)
        pairs = np.asarray(
[...]

PR #84, opened 2026-06-26
switch Molecular Comparison to KDTree

Bud-Macaulay:
As mentioned in:
 #83 

"Should" just be a free win and further align the algorithms used between different datatypes.
Happy to discuss further

### DIFF
--- a/src/pymatgen/core/molecule_structure_comparator.py
+++ b/src/pymatgen/core/molecule_structure_comparator.py
@@ -13,6 +13,10 @@ from __future__ import annotations
 import itertools
 from typing import TYPE_CHECKING
 
+from scipy.spatial import cKDTree
+import numpy as np
+
+
 from monty.json import MSONable
 
 from pymatgen.util.due import Doi, due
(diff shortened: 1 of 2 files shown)


Bud-Macaulay:
Fixes #83

shyuep:
🤖 **Automated PR review** (generated by Claude on behalf of @shyuep). Based purely on the GitHub diff — no code was executed.

Solid `O(N²)` → KDTree refactor of `_get_bonds`; the cap logic (priority / 1-3 / halogen self-bond) is preserved and pair ordering stays `(i<j)`, so `priority_bonds`/`bond_13` membership checks still work.

A few points before merge:

- **`max_cutoff=20` Å is a silent correctness assumption.** Any bond whose cap exceeds 20 Å (e.g. a large `priority_cap` on a widely separated priority pair) would now be dropped. Safe for normal covalent radii, but better to derive the cutoff from `max(covalent_radius)·2·(1+max_cap)` or at least document/guard it.
- **Import block will fail lint.** `from scipy.spatial import cKDTree` / `import numpy as np` are placed in their own group above the existing third-party imports — ruff/isort will reorder. Fold them into the existing import group.
- **No test added.** Please add a regression test on a multi-atom molecule asserting the returned bond list is identical to the previous implementation.

Note: CI has not run on this branch yet, so correctness is unverified by the test suite.

Bud-Macaulay:
The current test file tests/core/test_molecule_structure_comparator.py sufficiently ensures no regression on a multi atom molecule. 

The max_cutoff is now derived from the covalent_radii and the ruff imports should be fixedd

Bud-Macaulay:
I've also just taken the liberty to (as i believe) fix the broken test in the test file for this module (commit ff3bb6b1fde6c9a0902ba70b0218011db1d90a65)

I'm not sure how it broke, i guess there was a refactor in the QCOutput model that changed the response shape at some point. 

I've now switched the test to use ['initial_molecule"] and ["molecule_from_last_geometry"] rather than indexing.

shyuep:
🤖 **Automated PR review** (generated by Claude on behalf of @shyuep) — re-review after the latest commit.

**Correctness (KDTree refactor):** Looks correct. `max_cutoff = 2 * max_radius * (1 + max_cap)` is a safe upper bound on every per-pair cap, since `r_i + r_j ≤ 2·max_radius`, the chosen cap factor `≤ max_cap`, and the halogen self-bond `0.1` factor only shrinks the cap. So `tree.query_pairs(max_cutoff)` cannot drop any pair that the old O(N²) `itertools.combinations` path would have kept — the result set is identical while avoiding the quadratic distance loop. Index remapping via `atoms[i_local]` preserves `i<j` ordering, so `priority_bonds`/`bond_13` membership checks still match.

**Tests:** Good to see the previously `xfail`-ed `test_structural_change_in_geom_opt` restored and updated to the current `QCOutput.data` keys (`initial_molecule` / `molecule_from_last_geometry`) — re-establishes regression coverage for this path.

**Minor:** distances are vectorized but caps are still applied in a Python loop; fine at typical molecule sizes, vectorizable later if it ever shows up in profiling. Consider an early return for `n_atoms < 2`.

**CI:** No checks are currently reported on the branch — please confirm the test workflow runs green before merging.

shyuep:

Nice O(N²)→KDTree speedup for bond detection. `max_cutoff = 2*max_radius*(1+max_cap)` is a safe upper bound — the per-pair caps (including the halogen ×0.1 shrink) only ever reduce the threshold, so `query_pairs` returns a superset of true bonds and correctness is preserved.

- Bond ordering changes (sorted `query_pairs` vs `itertools.combinations`). Please confirm no downstream consumer relies on the previous ordering.
- The PR also un-`xfail`s `test_structural_change_in_geom_opt` and switches it to the new `QCOutput` keys (`initial_molecule`, `molecule_from_last_geometry`). That is an unrelated fix bundled in — ideally its own PR, but acceptable.
- **CI has not run** (no checks present). Please trigger/confirm green CI before merge.

shyuep:
Thanks!

PR #81, opened 2026-06-23
[codex] Read Vaspwave charge augmentation data

hheei:
## Summary
- Read VASP HDF5 charge augmentation occupancies from `/charge/aug_occupancies_*` in `Vaspwave.get_chgcar()`.
- Preserve augmentation data in the returned `Chgcar.data_aug` so `Chgcar.write_file()` can emit legacy CHGCAR augmentation sections.
- Add std and noncollinear fixture coverage for HDF5 `data_aug` mapping.

## Root Cause
`Vaspwave.get_chgcar()` only read `/charge/grid` and `/charge/charge`, then constructed `Chgcar` without `data_aug`. Legacy CHGCAR parsing and writing already supported augmentation occupancies, so the HDF5 path was the missing link.

## Validation
- `uv run --with h5py pytest tests/io/vasp/test_outputs.py::TestVaspwave::test_h2_ispin2_std_fixture_get_chgcar_matches_chgcar tests/io/vasp/test_outputs.py::TestVaspwave::test_h2_ncl_fixture_get_chgcar_matches_chgcar`
- commit hook: ruff check, ruff format check, mypy, codespell

PR #85, opened 2026-06-28
Update condition for HSE06 run type determination

kavanase:
Currently `pymatgen` requires `HFSCREEN` in VASP to be 0.2 to be classified as a HSE06 run:
https://github.com/materialsproject/pymatgen-core/blob/83d1b9d9d6f6399dcf2de6155a17d4590dbd0fa7/src/pymatgen/io/vasp/outputs.py#L878 

But VASP users often use 0.207 or 0.208 (https://pubs.acs.org/doi/10.1021/acsomega.4c03861, https://www.researchgate.net/post/How-can-I-calculate-the-bandgap-by-using-HSE06-in-vasp ...) as a more exact match of the original HSE screening parameter choice – though notably VASP recommends just 0.2 for HSE06. 
In practice from experience this makes minimal difference to electronic structure, but a small difference in HFSCREEN can add up to a significant energy difference over a large supercell.

This PR expands the matching criteria to classify a calculation as a HSE06 run, from exact match of `HFSCREEN` to 0.2, to between 0.2 and 0.21.

### DIFF
--- a/src/pymatgen/io/vasp/outputs.py
+++ b/src/pymatgen/io/vasp/outputs.py
@@ -6650,6 +6650,47 @@ class Vaspwave(Vasprun):
         data = self._read_hdf5_dataset(data_path, h5_file=h5_file)
         return self._validate_volumetric_dataset(grid, data, data_path)
 
+    def _read_charge_augmentation_data(self) -> dict[str, dict[int, np.ndarray]]:
+        """Read charge augmentation occupancies from ``vaspwave.h5``."""
+        data_aug: dict[str, dict[int, np.ndarray]] = {}
+
+        with self._open_hdf5_file() as h5_file:
+            if "/charge/charge" not in h5_file:
+                return data_aug
+
+            component_count = h5_file["/charge/charge"].shape[0]
+            data_keys: tuple[str, ...]
+            if component_count == 1:
+                data_keys = ("total",)
+            elif component_count == 2:
+                data_keys = ("total", "diff")
+            elif component_count == 4:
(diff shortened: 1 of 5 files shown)


shyuep:
I am not sure if we should do this. Ultimately, this is used in compatibility schemes. If the HFSCREEN is not 0.2, then there will be major complications with energy matching.

What we can do is to have a HSE_HFSCREEN_TOL and allow for some variation. It will default to a very small number to preserve current behavior. If someone wants to allow 0.207 as a HSE06 run, they can set it to 0.1 or something.

kavanase:
Ok! I have no strong feelings, just spotted it and knew that many historically used 0.207/0.208. Was thinking there might be some compatibility issues but didn't spot any from a quick scan.
Happy to do the `HSE_HFSCREEN_TOL` implementation or leave as is

shyuep:

The HSE06 widening from `math.isclose(HFSCREEN, 0.20)` to `0.2 <= HFSCREEN <= 0.21` is reasonable for capturing the standard screening value. Two notes:
- The bound is one-sided in practice: it accepts `[0.20, 0.21]` but rejects values marginally below 0.20 (e.g. 0.199) that the old `isclose` matched. If intended, fine; otherwise consider a symmetric tolerance.
- The 3 failing CI jobs are **not caused by this PR**. They are pre-existing environment failures: `TotalDos` has no `set_Debye_frequency` (phonopy API change) and a numpy `ufunc 'degrees'` TypeError in `_angle_dot`. Both are fixed in #81; the `run_type` assertion itself passes.

LGTM on the logic — recommend merging/rebasing on #81 so CI goes green.

PR #88, opened 2026-07-09
Change handling of standard cells in SpacegroupAnalyzer

### LINKED ISSUE #4679: get_primitive_standard_structure() incorrectly expands primitive hexagonal P cells from N to 3N sites when a ≈ c
### Python version

Python 3.12.3

### Pymatgen version

2025.6.14

### Operating system version

Ubuntu 24.04.4 LTS

### Current behavior

`SpacegroupAnalyzer.get_primitive_standard_structure()` returns 3N sites for a N-site primitive hexagonal P-6m2 structure.
The structure has nearly equal a and c lengths. A hard-coded equal-length heuristic in `get_conventional_standard_structure()` incorrectly treats it as a primitive rhombohedral representation and applies a determinant-3 supercell transformation.
The later primitive-standard logic still uses the original P-6m2 / hexagonal classification and therefore returns the expanded cell without reducing it.

### Expected Behavior

`get_primitive_standard_structure()` should return a symmetry-equivalent 3-site primitive structure.
The rhombohedral expansion should be gated by the existing rhombohedral space-group/lattice classification, or should also validate the lattice angles.

### Minimal example

```Python
from pymatgen.core import Structure
from pymatgen.symmetry.analyzer import SpacegroupAnalyzer

structure = Structure(
    lattice=[
          [4.0570938, 0.0, 0.0],
          [0.0, 4.0572993, 0.0],
          [0.0, -2.02864965, 3.51372426],
      ],
     species=["Gd", "Se", "F"],
     coords=[
          [0.67496701, 2.02864967, 1.17124141],
          [2.70478423, -2.02864966e-8, 2.34248285],
          [0.67734256, 0.0, 0.0],
      ],
     coords_are_cartesian=True,
  )

sga = SpacegroupAnalyzer(structure, symprec=0.01)

print(sga.get_space_group_symbol())
print(sga.get_lattice_type())
print(len(sga.find_primitive()))
print(len(sga.get_primitive_standard_structure()))
```

_No response_

Sanftperlig:
Fixes a variety of small issues with the SpacegroupAnalyzer, in particular with `get_primitive_standard_structure` and similar functions:
1. `get_primitive_standard_structure` now actually checks for the rhombohedral lattice type (instead of checking if the lattice is equal to the string "rhombohedral"). The unused code was changed to use .species instead of .specie so disordered structures work.
2. `get_primitive_standard_structure` now only creates the standard structure once, instead of twice (one extra time via `get_conventional_to_primitive_transformation_matrix`).
3. Centering checks of the space group now only check the first letter, as that is the only relevant letter.
4. All hexagonal space groups start with P, so checking the lattice type was duplicated.
5. `get_primitive_standard_structure` now always returns a float64-typed matrix.
6. Fixes materialsproject/pymatgen#4679 by adding an angle check to `get_conventional_standard_structure`.
7. Removes the list(range) in `itertools.permutations`, as that is unnecessary.
8. Fixes some Docstring typos and indentation mistakes (checked against VS Code tooltips).

## Checklist

### DIFF
--- a/src/pymatgen/core/structure.py
+++ b/src/pymatgen/core/structure.py
@@ -4741,7 +4741,7 @@ class Structure(IStructure, collections.abc.MutableSequence):
         """Rapidly construct common prototype structures.
 
         Args:
-            prototype: Name of prototype. e.g. cubic, rocksalt, perovksite etc.
+            prototype: Name of prototype. e.g. cubic, rocksalt, perovskite etc.
             species: List of species corresponding to symmetrically distinct sites.
             **kwargs: Lattice parameters, e.g. a = 3.0, b = 4, c = 5. Only the required lattice parameters need to be
                 specified. For example, if it is a cubic prototype, only a needs to be specified.
--- a/src/pymatgen/symmetry/analyzer.py
+++ b/src/pymatgen/symmetry/analyzer.py
@@ -699,9 +699,8 @@ class SpacegroupAnalyzer:
                 for idx in range(2):
                     transf[idx][sorted_dic[idx]["orig_index"]] = 1
                 c = lattice.abc[2]
-            elif self.get_space_group_symbol().startswith(
-                "A"
-            ):  # change to C-centering to match Setyawan/Curtarolo convention
+            # change to C-centering to match Setyawan/Curtarolo convention
+            elif self.get_space_group_symbol().startswith("A"):
                 transf[2] = [1, 0, 0]
                 a, b = sorted(lattice.abc[1:])
                 sorted_dic = sorted(
@@ -743,10 +742,13 @@ class SpacegroupAnalyzer:
             # check first if we have the refined structure shows a rhombohedral
             # cell
             # if so, make a supercell
-            a, b, c = lattice.abc
-            if np.all(np.abs([a - b, c - b, a - c]) < 0.001):
+            if np.allclose(lattice.abc, lattice.a, atol=1e-3, rtol=0.0) and np.allclose(
+                lattice.angles, lattice.alpha, atol=1e-2, rtol=0.0
+            ):
                 struct.make_supercell(((1, -1, 0), (0, 1, -1), (1, 1, 1)))
                 a, b, c = sorted(struct.lattice.abc)
+            else:
+                a, b, c = lattice.abc
 
             if abs(b - c) < 0.001:
                 a, c = c, a
@@ -755,7 +757,6 @@ class SpacegroupAnalyzer:
                 [a / 2, a * math.sqrt(3) / 2, 0],
                 [0, 0, c],
             ]
-            lattice = Lattice(new_matrix)
             transf = np.eye(3, dtype=np.float64)  # type:ignore[assignment]
 
         elif latt_type == "monoclinic":
--- a/src/pymatgen/symmetry/analyzer.py
+++ b/src/pymatgen/symmetry/analyzer.py
@@ -775,7 +775,7 @@ class SpacegroupAnalyzer:
                 b = sorted_dic[1]["length"]
                 c = lattice.abc[2]
                 new_matrix = None
-                for tp2 in itertools.permutations(list(range(2)), 2):
+                for tp2 in itertools.permutations(range(2), 2):
                     m = lattice.matrix
                     latt2 = Lattice([m[tp2[0]], m[tp2[1]], m[2]])
                     lengths = latt2.lengths
@@ -819,7 +819,7 @@ class SpacegroupAnalyzer:
                 # keep the ones with the non-90 angle=alpha and b<c
                 new_matrix = None
 
-                for tp3 in itertools.permutations(list(range(3)), 3):
+                for tp3 in itertools.permutations(range(3), 3):
                     m = lattice.matrix
                     a, b, c, alpha, beta, gamma = Lattice([m[tp3[0]], m[tp3[1]], m[tp3[2]]]).parameters
                     if alpha == 90 or b >= c:
--- a/src/pymatgen/symmetry/analyzer.py
+++ b/src/pymatgen/symmetry/analyzer.py
@@ -235,7 +235,7 @@ class SpacegroupAnalyzer:
 
     def get_lattice_type(self) -> LatticeType:
         """Get the lattice for the structure, e.g. (triclinic, orthorhombic, cubic,
-        etc.).This is the same as the crystal system with the exception of the
+        etc.). This is the same as the crystal system with the exception of the
         hexagonal/rhombohedral lattice.
 
         Raises:
@@ -533,8 +533,8 @@ class SpacegroupAnalyzer:
             conv_lattice = conv.lattice
         lattype = self.get_lattice_type()
 
-        if "P" in self.get_space_group_symbol() or lattype == "hexagonal":
-            return np.eye(3)
+        if self.get_space_group_symbol().startswith("P"):
+            return np.eye(3, dtype=np.float64)
 
         if lattype == "rhombohedral":
             # Check if the conventional representation is hexagonal or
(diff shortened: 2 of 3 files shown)


shyuep:
🤖 **Automated PR review by Claude** (posted via @shyuep). Based purely on static analysis of the diff and CI results — no local execution.

Nice, well-scoped cleanup. All CI green. Summary of assessment:

**Looks correct / good improvements**
- `startswith("P")` (and `"I"/"F"/"C"/"A"`) instead of substring `in` checks is more correct — centering is encoded only in the leading letter, so the old `in` tests were latent bugs. Dropping the separate `lattice == "hexagonal"` branch is safe since all hexagonal space groups are P-centered.
- `.specie` → `.species` correctly enables disordered structures.
- Threading `conv_lattice` through avoids building the conventional structure twice — good perf win.
- The #4679 fix (adding an angle check via `np.allclose(angles, alpha)` so a hexagonal cell with a≈c isn't misclassified as rhombohedral) is the right fix and is covered by the new regression test.
- Removing the duplicate `return Structure.from_sites(new_sites)` and the `math.radians` / `float64` dtype tidy-ups are all fine.

**Minor**
- In `get_conventional_standard_structure`, the trailing `lattice = Lattice(new_matrix)` was removed in the rhombohedral branch. Please confirm `lattice` (vs `new_matrix`) isn't consumed later in that branch — CI suggests it's fine, but worth a quick eyeball.
- The PR checklist leaves "Tests added" unchecked, though a targeted #4679 test is included. Consider also asserting the disordered-structure (`.species`) path, since that behavior change is otherwise untested.

Overall LGTM pending the minor confirmation above.

Sanftperlig:
`lattice` isn't read from in `get_conventional_standard_structure`, until it is overwritten by the new lattice, so that works.

shyuep:

Solid cleanup + perf improvement:
- Threading `conv_lattice` through `get_conventional_to_primitive_transformation_matrix` to avoid recomputing `get_conventional_standard_structure()` is a good win; call sites updated correctly.
- `"X" in symbol` → `symbol.startswith(...)` for centering checks is more correct (avoids false substring matches). `site.specie`→`site.species` and `math.pi*x/180`→`math.radians` are good fixes.
- Removing the duplicate early `return Structure.from_sites(new_sites)` in the rhombohedral branch is correct.

One question: the old code special-cased `lattice == "hexagonal"`; that's dropped in favor of `startswith("P")`. Hexagonal groups are P/R so this should be equivalent — please confirm an R-centered / hexagonal-setting case still yields the intended cell. CI is green.

Sanftperlig:
`lattice == "hexagonal"` checked if `get_lattice_type` returned hexagonal, e. g. if the space group number was between 143 and 194.
Among these groups only seven are non-primitive (the R... space groups), which are the seven explicitly changed from their crystal system to rhombohedral in `get_lattice_type`:
https://github.com/materialsproject/pymatgen-core/blob/c8cfee71f3ce5d92c5b5f6ca3d363f8f3a9f4495/src/pymatgen/symmetry/analyzer.py#L247-L251

The behaviour for the trigonal and hexagonal space groups before therefore was this: All except the seven rhombohedral groups have a space group str containing "P", therefore all of them pass. The remaining seven groups had lattice == "rhombohedral", so they don't pass the second condition either.

Now, that second condition was removed, as no space group can reach it and pass. In fact, replacing the lattice type check completely should also be fine (given those seven groups do start with "R").

I'm also going to remove the following `abs(lengths[0] - lengths[2]) < 1e-4` check, as when testing with `Structure.from_spacegroup`, no space group triggers this case. Unless there's another reason to keep it, I don't think it is useful.

Also, if a cell is A-centered, the transformation matrix would have to be `np.array([[2, 0, 0], [0, 1, 1], [0, -1, 1]], dtype=np.float64) / 2` instead of `np.array([[1, -1, 0], [1, 1, 0], [0, 0, 2]], dtype=np.float64) / 2`, right? Note that all four A-centered cells are orthorhombic. (I would strongly assume nonstandard centerings such as A-1 instead of P-1 are removed by standardization?) This fails some other tests (see CI), but I don't see why it would be wrong?

The final `return np.eye(3)` also should be unreachable, I'm replacing it with a raise (if it is better to fail silently, it should be changed back).

shyuep (commented):
🤖 **Automated PR review generated by Claude** (posted on behalf of @shyuep)

Solid cleanup overall — switching centering checks to `startswith` (first letter only, avoids spurious substring matches), `math.radians`, using `.species` so disordered structures work, float64 typing, and dropping the duplicate `return` / redundant `get_conventional_standard_structure` call are all genuine improvements. Tests included. 👍

Two blockers before merge:

**1. The CI failures look like real regressions, not flakiness.** The failing jobs (lowest-direct + highest/3.14) both fail in the A-centered / rhombohedral code paths this PR rewrote:
- `test_get_primitive_standard_structure`: `prim.lattice.angles` → `(140.69, …, 90.0)` vs expected `(90, …, 86)`.
- `TestKPathLatimerMunro.test_kpath_acentered`: k-point label set mismatch.

**2. Behaviour changes to double-check:**
- The new standalone `A`-centering branch returns `[[2,0,0],[0,1,1],[0,-1,1]]/2`, whereas `A` was previously handled by the `C` branch (including its monoclinic special-case matrix). Please confirm correctness for A-centered monoclinic — the failing `test_kpath_acentered` hints it may be off.
- `get_conventional_to_primitive_transformation_matrix` now raises `ValueError` for unrecognized space groups instead of falling back to `np.eye(3)`; stricter, but could surface on inputs that previously passed silently.

Please get the two failing tests green (or update expected values with justification) before merge.

Sanftperlig:
The reason why A-centering worked before (even though the matrix is only valid for C-centering, compare for example spglib [spacegroup.c line 343](https://github.com/spglib/spglib/blob/a6b561fb60cdd021a1ac90852c3c2ae14405b2c8/src/spacegroup.c#L344)) is that it was called after `get_conventional_standard_structure`, which converts A-centering to C-centering, but doesn't change the space group symbol.
Now the behaviour is the same as before (except for the bugfixes for rhombohedral structures).
The `international_monoclinic` parameter is also unused now, what should happen to it - is the signature allowed to change?

shyuep:
🤖 _Automated PR review generated by Claude, posted on behalf of @shyuep. Based purely on static analysis of the diff and CI results — no code was executed locally._

Solid refactor of the standardization routines. Highlights:
- Switching centering detection from substring `in` checks to `.startswith(...)` is correct and more robust (avoids false matches when a centering letter appears elsewhere in the symbol).
- Replacing the silent `return np.eye(3)` fallthrough with an explicit `raise ValueError` for unrecognized space groups is a good defensive change.
- `site.specie` → `site.species` in the rhombohedral branch is a genuine correctness fix, as are the `math.radians`/`np.allclose`/`math.isclose` tolerance cleanups.

Please confirm/add tests before merge:
- `get_conventional_to_primitive_transformation_matrix` now always returns the R→P matrix for `R` groups; the old length check (`|a−c|<1e-4 → eye(3)`) is gone. Add a regression test on a rhombohedral group in both rhombohedral- and hexagonal-setting conventional cells to lock in the intended behavior.
- A-centered handling now relies on `get_conventional_standard_structure` standardizing A→C first (per the new comment). A round-trip test on an A-centered group would guard against regressions.

CI is green. LGTM once the R/A-C centering cases are covered by tests.

Sanftperlig:
Note that the reliance of standardizing A-centering to C-centering is not new, it was simply not stated before. The Amm2-cell test from this branch also checks that an A-centered spacegroup uses the C-centering. I am changing it to also verify that the a-length is preserved in c (as expected from this standardization).
The current spacegroup 166 test covers a rhombohedral structure starting in the hexagonal setting. It correctly primitivizes to the rhombohedral primitive cell, but I will add a test to also check the way back (from rhombohedral to hexagonal).

shyuep:
Thanks for the detailed update!

PR #91, opened 2026-07-14
Add an explicit error for disordered sites in xyz export

Sanftperlig:
Exporting a .xyz with disordered sites currently returns `AttributeError: attr='specie' not found on Site. Did you mean: 'species'?`.
As .xyz doesn't support partial occupancy, the error `ValueError("xyz only supports ordered sites.")` has been added to explain the issue.
Also, some missing type annotations were added to the surrounding functions. 

## Checklist

### DIFF
--- a/src/pymatgen/io/xyz.py
+++ b/src/pymatgen/io/xyz.py
@@ -114,7 +114,7 @@ class XYZ:
         with zopen(filename, mode="rt", encoding="utf-8") as file:
             return cls.from_str(file.read())  # type:ignore[arg-type]
 
-    def as_dataframe(self):
+    def as_dataframe(self) -> pd.DataFrame:
         """Generate a coordinates data frame with columns: atom, x, y, and z
         In case of multiple frame XYZ, returns the last frame.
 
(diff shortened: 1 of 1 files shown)


shyuep:
🤖 _Automated PR review generated by Claude, posted on behalf of @shyuep. Based purely on static analysis of the diff and CI; no code was run locally._

The change is correct and well-scoped. Guarding `_frame_str` with `frame_mol.is_ordered` is the right place — it covers both `__str__` and `write_file`, since silently writing `site.specie` for a disordered site would otherwise emit meaningless output. The added return-type annotations are a nice touch, and CI (lint + tests across all platforms) is green.

Two minor, non-blocking suggestions:
- Add a unit test asserting `ValueError` is raised when exporting a disordered `Structure`/`Molecule`, so the behavior doesn't silently regress.
- `as_dataframe` isn't covered by the guard and will still fail with an opaque error on disordered input; consider raising the same explicit message there for consistency.

Looks good to merge once a regression test is added.

shyuep:
Thanks,

PR #92, opened 2026-07-17
Handle uncommon coordinate shapes

Sanftperlig:
Hardens the shape of `.coords`/`.frac_coords`/`.cart_coords` and derivatives:
1. `IStructure.frac_coords` and `IStructure.cart_coords` are now always (N, 3) shaped (and therefore 2D), even if N is 0. This is relevant for some applications, e. g. unpacking or numba specialization.
2. `Lattice.get_all_distances` (used by `Structure.distance_matrix`) now doesn't crash for siteless structures, but instead produces an empty array of the correct 2D-shape.
3. PeriodicSites/IStructures are better protected from misshapen coordinates: On `PeriodicSite` creation and on `.coords`/`.frac_coords` setting, the shape is checked to be (3,). This protects users of those values from recieving arbitrary coordinate shapes (e. g. (2,), (1, 3), (999, 999, 999)...).
4. Docstrings for lattice-associated functions were clarified. The `get_x_coords`-functions are regularly used by pymatgen with (N, 3)-shaped inputs, but the docstrings did not show this option. PeriodicSites now note more clearly that the expectation from a lattice change is that the fractional coordinate is supposed to remain unchanged.
5. In `PeriodicSite`, the input coordinates are now only copied if they are a) cartesian, b) converted to the unit cell or c) not a float64-array. This is slightly different from before, as before they were always copied (potentially again) if `skip_checks == False` and potentially copied a second (or third) time if the np.mod does not return float64.

## Checklist

### DIFF
--- a/src/pymatgen/core/lattice.py
+++ b/src/pymatgen/core/lattice.py
@@ -1619,8 +1619,10 @@ class Lattice(MSONable):
             2d array of Cartesian distances. E.g the distance between
             frac_coords1[i] and frac_coords2[j] is distances[i,j]
         """
-        _v, d2 = pbc_shortest_vectors(self, frac_coords1, frac_coords2, return_d2=True)
-        return np.sqrt(d2)
+        if len(frac_coords1) and len(frac_coords2):
+            _v, d2 = pbc_shortest_vectors(self, frac_coords1, frac_coords2, return_d2=True)
+            return np.sqrt(d2)
+        return np.empty((len(frac_coords1), len(frac_coords2)), dtype=np.float64)
 
     def is_hexagonal(
         self,
--- a/src/pymatgen/core/lattice.py
+++ b/src/pymatgen/core/lattice.py
@@ -215,7 +215,7 @@ class Lattice(MSONable):
         """Get the Cartesian coordinates given fractional coordinates.
 
         Args:
-            fractional_coords (3x1 array): Fractional coords.
+            fractional_coords (3x1 or Nx3 array): Fractional coords.
 
         Returns:
             Cartesian coordinates
@@ -226,7 +226,7 @@ class Lattice(MSONable):
         """Get the fractional coordinates given Cartesian coordinates.
 
         Args:
-            cart_coords (3x1 array): Cartesian coords.
+            cart_coords (3x1 or Nx3 array): Cartesian coords.
 
         Returns:
             Fractional coordinates.
@@ -248,7 +248,7 @@ class Lattice(MSONable):
         class in `pymatgen.analysis.ferroelectricity.polarization`.
 
         Args:
-            cart_coords (3x1 array): Cartesian coords.
+            cart_coords (3x1 or Nx3 array): Cartesian coords.
 
         Returns:
             Lattice coordinates.
--- a/src/pymatgen/core/sites.py
+++ b/src/pymatgen/core/sites.py
@@ -416,7 +416,7 @@ class PeriodicSite(Site, MSONable):
 
     @lattice.setter
     def lattice(self, lattice: Lattice) -> None:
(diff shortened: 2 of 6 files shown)


shyuep:

Solid hardening of coordinate shapes; CI is green on all platforms.

- The `(3,)` guards in `PeriodicSite.__init__` and the `coords`/`frac_coords` setters are a sensible defensive fix, and the empty `(N, 3)` handling in `Structure.cart_coords`/`frac_coords` and `Lattice.get_all_distances` correctly preserves 2D shape for siteless structures.
- Minor: the strict `(3,)` check is a small behavior change — any caller that previously passed a `(1, 3)` array to `PeriodicSite` will now raise. Low risk given the added tests, but worth a note in the release notes/changelog.
- `np.empty((N, 0))`/`(0, M)` returns are safe since they contain no uninitialized elements.

LGTM pending maintainer sign-off.

Sanftperlig:
On the comment suggesting `(1, 3)`-shaped arrays worked fine previously: As the format was simply never checked, *setting* the coordinate certainly worked, but any users were then exposed to this misshapen coordinate. In some cases, this happens to be fine, but generally it will cause issues up to exceptions.
As an example, try creating a Structure from the site and then printing the `__str__`:
```py
print(Structure.from_sites([PeriodicSite("H", np.zeros((1, 3)), Lattice.cubic(1))]))
```
This will cause a `TypeError` from the `to_str` function, as you cannot format a non-scalar value like this:
https://github.com/materialsproject/pymatgen-core/blob/488ad74cc5ecaba5d24c1726e2762fb47f31f5ef/src/pymatgen/core/structure.py#L1196-L1197
Other exceptions will be raised in other circumstances, e. g. if you tried to combine a `(1, 3)`-coordinate site with a normal `(3,)` site.

But even without trying to use the `PeriodicSite` in a `Structure`, this coordinate shape will cause problems: With `to_unit_cell=True`, the zip through all coordinates will fail (as the first dimension is of size 1, while the pbc tuple is of size 3 and the zip is strict):
https://github.com/materialsproject/pymatgen-core/blob/488ad74cc5ecaba5d24c1726e2762fb47f31f5ef/src/pymatgen/core/sites.py#L357-L358

Additionally, the `a`/`b`/`c`/`x`/`y`/`z` properties will be wrong (or impossible to access), as they index into the fractional or cartesian coordinates, either causing the `(1, 3)`-shaped array to be returned or an `IndexError`:
https://github.com/materialsproject/pymatgen-core/blob/488ad74cc5ecaba5d24c1726e2762fb47f31f5ef/src/pymatgen/core/sites.py#L442-L450
If you happened to only set `a` (or `x`), you'd cause a silent difference in behaviour: With the `(1, 3)`-shaped coordinate, `a` is equal to the entire array, therefore setting it to a float (after all, it is supposed to be one!), will set the entire array to this value, whereas with `(3,)` only the first value in the array changes, as intended:
```py
bad_site = PeriodicSite("H", np.zeros((1, 3)), Lattice.cubic(1))
bad_site.a = 0.5
assert np.all(bad_site.frac_coords == np.array([0.5, 0.5, 0.5]))

good_site = PeriodicSite("H", np.zeros(3), Lattice.cubic(1))
good_site.a = 0.5
assert np.all(good_site.frac_coords == np.array([0.5, 0, 0]))
```

I could go on with further situations where this unchecked size will cause exceptions, but I think the point is clear.

One more note regarding point 5: If the intended behaviour is to always copy the input coordinates (such that the PeriodicSite always has its own saved coordinates), it would be reasonable to change the `.asarray` to `.array` (thus always creating another copy), potentially with a `if frac_coords is coords` check to avoid extra copying.

shyuep:
Thanks!

PR #93, opened 2026-07-17
Use math.radians, math.degrees (or np.rad2deg, np.deg2rad for arrays) instead of calculating 180.0 / pi or pi / 180.0

Sanftperlig:
Changes all instances (that I could find) of manual radians <-> degrees conversion to the math function (or the numpy function, if needed).

### DIFF
--- a/src/pymatgen/core/interface.py
+++ b/src/pymatgen/core/interface.py
@@ -853,7 +853,7 @@ class GrainBoundaryGenerator:
 
         # For rhombohedral system, ratio = (1 + 2 * cos(alpha)) / cos(alpha)
         elif lat_type == "r":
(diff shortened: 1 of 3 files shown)


shyuep:

Clean, mechanical refactor replacing manual `*pi/180` / `*180/pi` with `math.radians`/`math.degrees` for scalars and `np.rad2deg` for arrays. Correctly keeps the NumPy variants where inputs are vectorized (`Lattice.angles`, `get_angles`). CI green.

- Values are mathematically equivalent; only sub-ULP float differences vs. the old expressions, so no behavior change expected.
- Good catch on the follow-up commit fixing `get_angles` to use the array-safe conversion.

LGTM.

PR #95, opened 2026-07-20
Fix `Outcar` parsing OUTCAR with truncated table

### LINKED ISSUE #4550: Hang when parsing OUTCAR
### Python version

Python 3.12.11

### Pymatgen version

master

### Operating system version

Gentoo

### Current behavior

Pymatgen never ends parsing the attached OUTCAR.

[OUTCAR.gz](https://github.com/user-attachments/files/23653430/OUTCAR.gz)

### Expected Behavior

Parsing finishes. Maybe throws an error. Not hangs.

### Minimal example

```python
from pymatgen.io.vasp.outputs import Outcar
o=Outcar('OUTCAR')
```

_No response_

DanielYang59:
fix https://github.com/materialsproject/pymatgen/issues/4550


DanielYang59:
@yantar92 it turns out your OUTCAR is truncated (only 140 kpoints), it should be fixed now. 

Can I have your permission to add that OUTCAR as a test file to pmg-core?

yantar92:
Haoyu YANG 杨浩宇 ***@***.***> writes:

> @yantar92 it turns out your OUTCAR is truncated (only 140 kpoints), it should be fixed now. 
>
> Can I have your permission to add that OUTCAR as a test file?

Feel free.

shyuep:

Good defensive fix — guarding against a truncated table that would otherwise trigger catastrophic regex backtracking / a hang in `read_table_pattern`. The early header/footer check with a warning and empty return is clean, and the truncated-OUTCAR fixture + test cover the case. CI is green.

Minor notes:
- When the header is entirely absent it returns `[]` silently (no warning); that matches the prior no-match behavior, so acceptable.
- `header_pattern`/`footer_pattern` are re-searched per call; negligible here, but could be precompiled if this path ever gets hot.

Nice catch on the hang. LGTM.

shyuep:

**Scope:** Guards `read_table_pattern` against truncated tables (header present, footer missing) that previously caused the parser to hang (fixes #4550). +15 LOC, plus a regression test and an `OUTCAR.truncated.gz` fixture. CI is green.

**Looks good**
- Targeted, defensive fix: early-return with `warnings.warn(..., stacklevel=2)` and empty result instead of falling into catastrophic regex backtracking.
- When `attribute_name` is set, it stores `[]` into `self.data[attribute_name]` — consistent with the normal-path contract, so downstream `outcar.data[...]` access stays valid.
- Good test coverage: asserts both the `UserWarning` and the empty result.

**Minor (non-blocking)**
- Adds two extra full-text `re.search` passes (header + footer, `DOTALL`) on every `read_table_pattern` call. Negligible in practice, but on very large OUTCARs it is a small constant overhead on the happy path.
- Header-absent case now returns `[]` silently (warning only fires when a header is found). This matches prior effective behavior, so fine — just noting the asymmetry.

**Verdict:** LGTM — well-scoped fix with a regression test. Good to merge.

PR #96, opened 2026-07-20
Ask for permission for using uploaded files in issue template

DanielYang59:
I'm by no means a lawyer, but I think users retain ownership of their uploaded files and pmg might not automatically have the right to add their files to the test suite just because users provided a file during bug report? Previously I have to manually ask the user for permission which was a bit tedious


shyuep:

Sensible, low-risk change — adds an explicit licensing checkbox to the bug-report template so maintainers can add uploaded reproduction files to the test suite with clear permission. Non-code change; lint passes.

Nit: the checkbox is intentionally optional (no `required`), which is correct since it only applies when a file is actually uploaded. Wording is clear. LGTM.

PR #97, opened 2026-07-20
Setters for `Structure.frac_coords` and `Structure.cart_coords`

Sanftperlig:
Currently, changing all coordinates in a Structure requires either manually looping over all sites, or using one of the functions made for specific tasks. One example for this is a simple translation of all sites by a constant (vector), which `translate_sites` can be used for. But that requires an iterable of all site indices (or, to follow the previous typing: a sequence of them), which is unnecessary, if all sites are modified.

Assuming no PBC is needed, this change simplifies such a translation to the following intuitive format:
```py
structure.cart_coords += vector
```

In the same way, more complex modifications of the whole coordinate array are enabled.
Once #50 is merged, this can be simplified further (with a less insignificant performance benefit than right now), as no looping would be needed (given the entire array can then be accessed at once). This change means the coordinate arrays act even more like normal arrays.

The setters check for an equal length between the set coordinates and the new coordinates, as I doubt anyone would want to intentionally pass a smaller coordinate array to only modify the first coordinates (and longer coordinates would remain unused).
The per-coordinate validation from #92 should also work with this change.

## Checklist

### DIFF
--- a/src/pymatgen/core/structure.py
+++ b/src/pymatgen/core/structure.py
@@ -4031,12 +4031,22 @@ class Structure(IStructure, collections.abc.MutableSequence):
     @IStructure.frac_coords.setter
     def frac_coords(self, fractional_coords: ArrayLike) -> None:
         """Sets the fractional coordinates of the Sites in the Structure."""
+        if len(fractional_coords) != len(self.sites):
+            raise ValueError(
+                "A Structure must have as many coordinates as it has sites, "
+                f"{len(fractional_coords)=} != {len(self.sites)=}."
+            )
         for i, site in enumerate(self.sites):
             site.frac_coords = np.asarray(fractional_coords[i], dtype=np.float64)
 
     @IStructure.cart_coords.setter
     def cart_coords(self, cartesian_coords: ArrayLike) -> None:
         """Sets the cartesian coordinates of the Sites in the Structure."""
+        if len(cartesian_coords) != len(self.sites):
+            raise ValueError(
+                "A Structure must have as many coordinates as it has sites, "
+                f"{len(cartesian_coords)=} != {len(self.sites)=}."
+            )
         for i, site in enumerate(self.sites):
             site.coords = np.asarray(cartesian_coords[i], dtype=np.float64)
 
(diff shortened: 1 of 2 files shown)


shyuep:

Clean addition — extending the `IStructure` coordinate properties with `Structure`-level setters is the right pattern, and the `structure.cart_coords += vector` idiom reads well and is covered by the new test. CI is green.

Two minor notes:
- The setters validate the outer length but not that each row has 3 components. You note per-element validation lands with #92, so fine to defer.
- `translate_sites` now takes `Iterable[int]` instead of `Sequence[int]`. If the body ever indexes `indices` or iterates it more than once, a one-shot iterable (e.g. a generator) would break — worth confirming it's consumed exactly once.

Otherwise looks good.

shyuep:
Thanks for the contribution. But I believe IStructure is meant to be immutable. So the setters should be in Structure instead? Thanks.

Sanftperlig:
Those setters are in `Structure`, but the decorator is `@IStructure.frac_coords.setter`, because the `frac_coords` property is a property of the superclass `IStructure`. This way, the `frac_coords` property doesn't need to be redefined, but only the subclass has the setter.
The other option is to do it like the `lattice` property, which is overridden by `Structure`:
In `IStructure`, the property is defined as https://github.com/materialsproject/pymatgen-core/blob/5afda6b0a2f8fbfec8f4156f347bcf0d63ba4494/src/pymatgen/core/structure.py#L1548-L1551 and then it is overridden by `Structure`, which also adds a setter: https://github.com/materialsproject/pymatgen-core/blob/5afda6b0a2f8fbfec8f4156f347bcf0d63ba4494/src/pymatgen/core/structure.py#L4035-L4046
As the property is created (overridden) in `Structure`, the decorator is simply `@lattice.setter`.

This means the property is duplicated (also duplicating the docstring, which differs here), even though it is exactly the same. `@IStructure.frac_coords.setter` avoids this, while the usage should be identical to `lattice`.

Sanftperlig:
@shyuep sorry for the @, but since this PR has been inactive for over a week (and since I don't want the PR to slowly become stale), I'm interested if there's anything blocking it.
I used the same kind of superclass-referencing property getter/setter decorator in #108, so I assume it is fine to use.
https://github.com/materialsproject/pymatgen-core/blob/39c044caa68d18af47318011464845a17707484a/src/pymatgen/core/structure.py#L3979-L3982

If there's a different problem with this PR, please tell me.

shyuep:
Thanks for the reminder and clarification. I will merge now.

PR #98, opened 2026-07-21
Add types for nwchem and vasp outputs

DanielYang59:
Add types for `nwchem` and vasp outputs

### DIFF
--- a/src/pymatgen/io/nwchem.py
+++ b/src/pymatgen/io/nwchem.py
@@ -23,7 +23,7 @@ from __future__ import annotations
 import os
 import re
 import warnings
-from collections.abc import Mapping
+from collections.abc import Mapping, Sequence
 from string import Template
 from typing import TYPE_CHECKING, Literal
 
@@ -35,8 +35,9 @@ from pymatgen.core.structure import Molecule, Structure
 from pymatgen.core.units import Energy, FloatWithUnit
 
 if TYPE_CHECKING:
-    from pathlib import Path
     from typing import Any, ClassVar, Self
+
+    from pymatgen.util.typing import PathLike
 
 NWCHEM_BASIS_LIBRARY: set | None = None
 if os.getenv("NWCHEM_BASIS_LIBRARY"):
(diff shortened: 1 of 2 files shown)


src/pymatgen/io/nwchem.py
```diff
@@ -336,19 +377,17 @@ class NwInput(MSONable):
...
+        tasks: list[NwTask],
+        directives: list[tuple[str, ...]] | None = None,
+        geometry_options: Sequence[str] = ("units", "angstroms"),
+        symmetry_options: Sequence[str] | None = None,
+        memory_options: str | None = None,
+    ) -> None:
         """Initialize a NwInput.
 
         Args:
-            mol: Input molecule. If molecule is a single string, it is used as a
```
DanielYang59, line 349: likely this docstring was copied from `GaussianInput` and I don't think mol could be a str here:
```python
out.extend(f" {site.specie.symbol} {site.x} {site.y} {site.z}" for site in self._mol)
```

shyuep:

**Scope:** Type annotations for `io/nwchem.py` (+126/-76) and `io/vasp/outputs.py` (+32/-32). No intended logic changes; CI (lint + full test matrix) is green.

**Looks good**
- Clean, comprehensive annotations. `TypeAlias` definitions for `NwTaskTheory`/`NwTaskOperation` literals are a nice touch and self-document valid values.
- `NwOutput` dunders (`__iter__`/`__getitem__`/`__len__`) are relocated, not duplicated — verified no double definition.

**Minor (non-blocking)**
- `NwTask.__init__` now types `charge: float` (non-optional), but the docstring still says "If None, charge on molecule is used." Either widen to `float | None` or drop the None wording. `from_molecule` correctly uses `float | None`.
- `NwTask.__init__` types `basis_set: dict[str, str]`, while `from_molecule` allows `dict[str, str] | str = "6-31g"`. The constructor annotation may be too narrow if a bare string is ever passed through.
- Not introduced here, but now more visible: `er_ev` in `read_igpar` does `np.array(map(float, match.groups()[1:4])) / 2` (a `map` object yields a 0-d object array). Worth a separate follow-up fix.

**Verdict:** LGTM — safe, low-risk typing PR. Optional to address the docstring/annotation nits before merge.

shyuep:

**Re-review after latest commit (`drop None in docstring`).** CI green (lint + full matrix).

- ✅ Addressed the earlier nit: the `NwTask.__init__` docstring no longer says "If None…" for `charge`, now consistent with the non-optional `charge: float` annotation.
- Minor (non-blocking, unchanged): `__init__` still types `basis_set: dict[str, str]` while `from_molecule` accepts `dict[str, str] | str`. Fine if the constructor is treated as the low-level API, but worth a comment noting strings must be resolved before construction.
- Follow-up (pre-existing, out of scope): `read_igpar` `np.array(map(float, …)) / 2` builds a 0-d object array — separate fix.

**Verdict:** LGTM — low-risk typing PR, safe to merge.

shyuep:
Thanks

PR #99, opened 2026-07-24
[Breaking] `Elfcar` and `Locpot` use proper spin_up/down instead of total and diff

### LINKED ISSUE #4307: ELFCAR data field names do not make sense
### Python version

N/A

### Pymatgen version

2025.2.18

### Operating system version

_No response_

### Current behavior

As @mkhorton has aptly noted:

https://github.com/materialsproject/pymatgen/blob/b5d03d7e313c4933ccad27ac0037cba4a1538351/src/pymatgen/io/vasp/outputs.py#L3947
https://github.com/materialsproject/pymatgen/blob/b5d03d7e313c4933ccad27ac0037cba4a1538351/src/pymatgen/io/vasp/outputs.py#L3973-L3976

This is pretty self-explanatory. The keys "total" and "diff" are not intuitive or appropriate names for spin up or spin down, especially when we have dedicated keys for those used throughout Pymatgen.

I am not making the change (yet) because I am not 100% certain that the current description is accurate. I assume it is based on the docstring, but I'd like to actually confirm this.

### Expected Behavior

See above

### Minimal example

```Python

```

_No response_

DanielYang59:
fix https://github.com/materialsproject/pymatgen/issues/4307

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -6,6 +6,12 @@ This file provides guidance to coding agents working in this repository.
 
 This is `pymatgen-core` — the trimmed core of the pymatgen ecosystem (the larger umbrella `pymatgen` package depends on this). It implements the foundational data structures (`Element`, `Site`, `Molecule`, `Structure`, `Lattice`, `Composition`), symmetry/lattice operations, the I/O layer for many DFT/MD codes (VASP, ABINIT, CIF, Gaussian, LAMMPS, CP2K, Q-Chem, …), and analysis modules. Python ≥ 3.11, src layout under `src/pymatgen/`.
 
+## Language
+
+Always communicate in **English**: replies, comments, PR/issue descriptions, code comments, and docstrings. This is an international open-source project — never write in another language, regardless of what language the user or another contributor uses.
+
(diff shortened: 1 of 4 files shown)


AGENTS.md
```diff
@@ -6,6 +6,12 @@ This file provides guidance to coding agents working in this repository.
 
 This is `pymatgen-core` — the trimmed core of the pymatgen ecosystem (the larger umbrella `pymatgen` package depends on this). It implements the foundational data structures (`Element`, `Site`, `Molecule`, `Structure`, `Lattice`, `Composition`), symmetry/lattice operations, the I/O layer for many DFT/MD codes (VASP, ABINIT, CIF, Gaussian, LAMMPS, CP2K, Q-Chem, …), and analysis modules. Python ≥ 3.11, src layout under `src/pymatgen/`.
 
+## Language
```
DanielYang59, line 9: Apparently out of scope for this PR, just quickly add it here because I'm seeing agents using languages other than English (e.g. c8c2f0257c0a7028f3b0524535bb96b703d94837)

DanielYang59:
I would mark it as ready now but I'm slightly concerned where old code tries to read new serialized data. But I would assume such breakage cannot really to avoided for such a key rename (similar situation for https://github.com/materialsproject/pymatgen-core/pull/94)...

shyuep:
Thanks all.

PR #101, opened 2026-07-24
Overhaul `get_symmetrically_distinct_miller_indices`

Sanftperlig:
This PR changes `get_symmetrically_distinct_miller_indices` from the ground up to improve it in various ways:
1. `get_symmetrically_distinct_miller_indices` now cleanly and transparently interacts with centered cells. Before, centered cells were primitivized based on the symmetry of the input structure (assumed to be standard, see 2.), except for trigonal structures (see a)). The new `cell` variable allows choosing between different options for this behaviour (see b)). **This changes the default behaviour for trigonal cells** (see c)).
2. Sorting of the indices is now completely deterministic, even against changes to the generator of the full hkl list. In the same way, the sorting has been slightly changed: Indices with smaller maximum absolute h/k/l still come first as before (which is relevant, compare materialsproject/pymatgen#2949), but then the indices with lower sums of absolute h/k/l follow, e. g. instead of [(1, 1, 1), (1, 1, 0), (1, 0, 0)] the cubic indices for max 1 are sorted as [(1, 0, 0), (1, 1, 0), (1, 1, 1)]. If both of those are the same (e. g. (0, 1, 1) and (1, 0, 1)), the larger `h`, then `k`, then `l` values are preferred. Previously, the implicit ordering was smaller maximum absolute h/k/l, then larger `h`, then `k`, then `l` (except for rhombohedral structures, or if a higher-order version of the plane happened to be in the set).
3. With the change in sorting, a potential change in returned `hkl` is possible, the current implementation prefers the value that comes first according to the above index. It was chosen to generally prefer positive indices and to sort the values in a more d_hkl following way. It would be possible to add `-h-k-l` as a third key if positive indices should be preferred even more. Alternatively, d_hkl could be used as the key (but then the pattern is less obvious).
4. A crystal family function was added to the `SpacegroupAnalyzer` (as the hexagonal family are the spacegroups for which hkil is interesting).
5. The typing is more specific and accurately types the return based on if hkil or hkl are requested.
6. If hkil are requested, but unlikely to be useful, a warning is emitted. I am open to removing this warning or changing the conditions, if the current assumptions are suboptimal.
7. Multiple changes were made to improve the performance as well:

- Miller indices with gcd above 1 are automatically excluded, as higher multiples of the same index were previously reduced down to the gcd (with some reliance on int=round)

- Previously, if a trigonal crystal system was detected, the SGA was created again to get a standard structure: https://github.com/materialsproject/pymatgen-core/blob/00104997ce2711c004cb848f56ee04f93a276c66/src/pymatgen/core/surface.py#L2081-L2086 Now, the amount of SGAs created is limited to the necessary amount.
- Previously, the final results list was created twice to handle the trigonal case separately, but the second list remained unused. Now, the trigonal case does not require special handling anymore.

a) Trigonal structures were previously separated out to get the reciprocal symmetry operations from the primitive cell. This actually only affected cells from the space groups with rhombohedral lattices in the expected way, because only they have a non-eye `get_conventional_to_primitive_transformation_matrix()`. As a side-effect the symmetry operations actually apply to the standard structure, meaning a nonstandard input structure returned either indices relating to its own symmetry (non-trigonal cell) or to the conventional standard cell (for the trigonal cells). This behaviour was not explicitly mentioned, see c) for how it is now.

b) Now, instead of the behaviour mentioned in a), there are five options. Note that there is no option that will return the exactly same result as before.
1. `input`: Use the lattice as-is, with all of its (non-spacegroup and centering) (a)symmetries. This is most similar to the previous behaviour, I would even assert that the returned sets of lattice planes are the same (but sometimes the representing lattice plane will be different, e. g. if (2, -1, 2) and (1, 1, 2) are symmetric, now (1, 1, 2) will be the representative instead of (2, -1, 2)). This is why it is the default behaviour. The key difference is trigonal space groups, especially the ones with rhombohedral lattice type. If given in their hexagonal cell, they will return the symmetry for that setting, not the primitive one. Simply transforming the cell beforehand does not do what one may expect, as then the indices belong to the primitive cell (compare `primitive` mode and `conv_np` mode).
2. `conventional`: Use the conventional standard cell to obtain the symmetry operations. For the trigonal cells, this is the closest to the previous behaviour. If the cell is centered, the symmetry operations of the primitive cell are used instead, but the indices apply to the conventional standard cell.
3. `primitive`: Use the primitive standard cell. The symmetry operations and the returned indices thus apply to that cell.
4. `conv_np`: Use the conventional standard cell, and do not use the operations from the primitive cell. It thus differs from `conventional` for centered cells.
5. `prim_2conv`: Use the primitive standard cell (like `primitive`), but convert the indices to indices in the conventional standard cell.

c) The default behaviour change cannot be avoided if we want to fix the behaviour difference between trigonal and all other cells, while not creating a too complex interface. As stated above, `input` is the current standard, even though I personally would choose `conventional`, as that matches the intent of the function more closely. But I believe more people will use the function with non-trigonal cells than trigonal cells, so the default was set to `input`.

I am open to name changes for the options, the current options try to guide users towards the behaviours they probably want (`input`, `conventional`, `primitive`) while adding other options for those with specific requirements.

## Checklist

### DIFF
--- a/src/pymatgen/core/lattice.py
+++ b/src/pymatgen/core/lattice.py
@@ -1740,7 +1740,7 @@ class Lattice(MSONable):
         to be used for hkl transformations.
 
         Args:
-            symprec: default is 0.001.
+            symprec: default is 0.01.
         """
         recp_lattice = self.reciprocal_lattice_crystallographic
         # Get symmetry operations from input conventional unit cell
--- a/src/pymatgen/core/structure.py
+++ b/src/pymatgen/core/structure.py
@@ -1052,9 +1052,9 @@ class IStructure(SiteCollection, MSONable):
                 ii. List of dict of elements/species and occupancies, e.g.
                     [{"Fe" : 0.5, "Mn":0.5}, ...]. This allows the setup of
                     disordered structures.
-            coords (Nx3 array): list of fractional/Cartesian coordinates of
+            coords (Nx3 array): Array of fractional/Cartesian coordinates of
                 each species.
-            charge (int): overall charge of the structure. Defaults to behavior
+            charge (int): Overall charge of the structure. Defaults to behavior
                 in SiteCollection where total charge is the sum of the oxidation
                 states.
             validate_proximity (bool): Whether to check if there are sites
@@ -3922,9 +3922,9 @@ class Structure(IStructure, collections.abc.MutableSequence):
                 ii. List of dict of elements/species and occupancies, e.g.
                     [{"Fe" : 0.5, "Mn":0.5}, ...]. This allows the setup of
                     disordered structures.
-            coords (Nx3 array): list of fractional/cartesian coordinates of
+            coords (Nx3 array): Array of fractional/cartesian coordinates of
                 each species.
-            charge (float): overall charge of the structure. Defaults to behavior
+            charge (float): Overall charge of the structure. Defaults to behavior
                 in SiteCollection where total charge is the sum of the oxidation
                 states.
             validate_proximity (bool): Whether to check if there are sites
--- a/src/pymatgen/core/surface.py
+++ b/src/pymatgen/core/surface.py
@@ -2120,12 +2120,19 @@ def get_symmetrically_distinct_miller_indices(
             with indices to the same maximum absolute index being sorted by ascending absolute sum over hkl.
     """
 
-    def hkl_sort(hkl: tuple[int, int, int]):
-        """Returns a deterministic sorting key for hkl lists."""
-        h = abs(hkl[0])
-        k = abs(hkl[1])
-        L = abs(hkl[2])
-        return (max(h, k, L), h + k + L, hkl)
+    def hkl_key(hkl: tuple[int, int, int]) -> tuple[int, int, int, int, int]:
+        """Returns a deterministic sorting key for hkl tuples.
+
+        This key was chosen to prefer smaller absolute maximum values and
+        First: Lower maximum absolute index (e. g. (100) over (200)).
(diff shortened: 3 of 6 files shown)


shyuep:
🤖 **Automated PR review generated by Claude** (posted on behalf of @shyuep). Not a substitute for human review.

Solid, well-documented overhaul. Deterministic `hkl_key` sorting, the `cell` mode abstraction, the new `get_crystal_family`/`CrystalFamily` alias, and the added overloads all improve clarity and typing. Performance cleanups (single SGA reuse, dropping non-reduced hkl via `gcd`, removing the duplicate results list) look correct. CI is green.

A few points:

1. **Docstring default mismatch**: the `cell` docstring reads `Literal[...] = "conventional"`, but the signature default is `"input"`. Please align to avoid confusion.
2. **Breaking behavior change** — worth an explicit callout in release notes: (a) default trigonal behavior changes, and (b) the new sort order and representative-index selection differ from before. Downstream consumers that assume a specific ordering or representative (e.g. `SlabGenerator`/`generate_all_slabs`, and cached results in dependent packages) could be affected. Recommend confirming those call sites are order-insensitive.
3. The primitive→conventional op conversion with `np.rint` + `atol=1e-8` and a `ValueError` guard is a reasonable safeguard; consider a targeted test covering an R-centered (rhombohedral) cell in the `conventional` path to lock in that branch.
4. Minor: on the `Literal[True] = True` overload the default is redundant but harmless.

No blocking issues from static review. Nice work.

PR #102, opened 2026-07-26
Fix BSPlotterProjected with disabled orbital summation

ShengLin1001:
## Summary

- avoid checking membership in `sum_morbs` when orbital summation is disabled
- add a regression test for selecting one individual orbital with `sum_morbs=None`

## Root cause

For a single multi-character orbital such as `dxy`, `_Orbitals_SumOrbitals` evaluated `elt in sum_morbs` before handling the documented `None` value. This raised `TypeError` even though `None` disables orbital summation.

This resubmits the fix from materialsproject/pymatgen#4167 after `electronic_structure` moved to `pymatgen-core`.

## Validation

- `pytest tests/electronic_structure/test_plotter.py::TestBSPlotterProjected -n 0` (2 passed)
- `ruff check src/pymatgen/electronic_structure/plotter.py tests/electronic_structure/test_plotter.py`
- `ruff format --check src/pymatgen/electronic_structure/plotter.py tests/electronic_structure/test_plotter.py`


shyuep:
Thanks!

PR #105, opened 2026-07-27
Allow smaller primitive Slabs, change the scale factor

Sanftperlig:
Introduces two key changes (and one smaller group of changes, explanation below):
1. New `allow_smaller_than_ouc` keyword for `SlabGenerator`: Disables lattice constraint and enables structure reduction for slab and OUC primitivization (effectively an enhanced reversion of #24 as an option).
2. The `scale_factor` of slabs is now recalculated based on the slab itself, not the scale factor generated from the miller index. **This changes the returned scale factor**: It is now slightly off-integer in most cases, c is now the entire slab c (not just the non-vacuum part - I am open to changing this back) and with primitivization the scale factor can now be non-integer.
3. Some small bugs, typing and performance cases were improved. E. g. if the primitivization is rejected, the energy remains unchanged; the OUC keeps its type; less structures etc. are copied since pymatgen generally trusts the user to not unknowingly modify the data in many other cases.

## Explanation

Starting with #24, the size of the surface of Slabs from a `SlabGenerator` was locked to the size of their non-reduced OUC.
This may be the expected behaviour in some use cases, but it blocks primitivizations which `Structure.get_primitive_structure` is only able to find for selected terminations, if the primitivization is not found for the OUC termination (and with `reduce=False` proper removal of centering is blocked).
One such example is the (104) surface of calcite (space group R-3c): Without primitivization, pymatgen generates the [010] x [-401] surface (which is fine, but very large). The common side vectors you will find in the literature are [42-1] x [010] (or equivalents), with [42-1] being reduced to a third of the full vector due to the centering (I will now call this [42-1]/3). This is the preferred surface representation because it is both smaller and orthagonal (gamma 90°, approx. 35 Å² surface area).
But with primitivization, pymatgen currently creates a [81-2]/3 x [-401] (15.6 Å x 24.2 Å, gamma 174.7°) slab with that same surface area, because `struct.get_primitive_structure(tolerance=tol, reduce=False)` returns that surface.
The new `allow_smaller_than_ouc` setting for the `SlabGenerator` (and for `get_all_slabs`) enables reduction of the primitive slab and primitive OUC (like before #24), but it also completely removes the lattice constraints on the primitive OUC and the following check that the OUC and slab have similar `a` and `b` (note that `gamma` was and still is missing from this surface equivalence check, but at least one test is based on gamma changing and I did not change this behaviour).
This way, `Structure.get_primitive_structure` fails at finding the [42-1]/3 x [010] for the OUC, but it manages to find [-4-21]/3 x [010] for a termination from `gen_possible_terminations` such as 0.125.

This is similar to before #24, as pymatgen used to find the [-4-21]/3 x [010], as the behaviour always followed the current `reduce=True`.

I also changed the `scale_factor` for the following reason: For a non-primitive slab, the [uvw] directions of the slab vectors were always available in the `Slab.scale_factor` ([0] for `a` and so on), with the matrix being `slab_lattice_matrix = slab.scale_factor @ structure.lattice.matrix` (correct for a and b; c only if no vacuum was used). The moment `primitive=True` is enabled, this is not guaranteed to be true anymore, even though there is no indication for this behaviour difference. The returned vectors remained the non-primitive vectors. The same holds for LLL reduction.
I changed this such that `slab_lattice_matrix = slab.scale_factor @ structure.lattice.matrix` always holds. This means that the scale factor changes, as stated above. In the primitive case, the side vectors stated by the scale factor now actually correspond to the side vectors of the slab. This does also change the c vector to be larger if vacuum is used and due to floating-point inaccuracy it causes all values to be slightly off-integer even for a non-primitive case. (This can be fixed by rounding if all values are nearly integer, if necessary.)

## Checklist

### DIFF
--- a/src/pymatgen/core/lattice.py
+++ b/src/pymatgen/core/lattice.py
@@ -1011,19 +1011,19 @@ class Lattice(MSONable):
 
         Returns:
             tuple[Lattice, NDArray[np.float_], NDArray[np.float_]]: (aligned_lattice, rotation_matrix, scale_matrix)
-            if a mapping is found. aligned_lattice is a rotated version of other_lattice that
-            has the same lattice parameters, but which is aligned in the
-            coordinate system of this lattice so that translational points
-            match up in 3D. rotation_matrix is the rotation that has to be
-            applied to other_lattice to obtain aligned_lattice, i.e.,
-            aligned_matrix = np.inner(other_lattice, rotation_matrix) and
-            op = SymmOp.from_rotation_and_translation(rotation_matrix)
-            aligned_matrix = op.operate_multi(latt.matrix)
-            Finally, scale_matrix is the integer matrix that expresses
-            aligned_matrix as a linear combination of this
-            lattice, i.e., aligned_matrix = np.dot(scale_matrix, self.matrix)
+                if a mapping is found. aligned_lattice is a rotated version of other_lattice that
+                has the same lattice parameters, but which is aligned in the
+                coordinate system of this lattice so that translational points
+                match up in 3D. rotation_matrix is the rotation that has to be
+                applied to other_lattice to obtain aligned_lattice, i.e.,
+                aligned_matrix = np.inner(other_lattice, rotation_matrix) and
+                op = SymmOp.from_rotation_and_translation(rotation_matrix)
+                aligned_matrix = op.operate_multi(latt.matrix)
+                Finally, scale_matrix is the integer matrix that expresses
+                aligned_matrix as a linear combination of this
+                lattice, i.e., aligned_matrix = np.dot(scale_matrix, self.matrix)
 
-            None is returned if no matches are found.
+                None is returned if no matches are found.
         """
         return next(
             self.find_all_mappings(other_lattice, ltol, atol, skip_rotation_matrix),
--- a/src/pymatgen/core/surface.py
+++ b/src/pymatgen/core/surface.py
@@ -1201,6 +1201,7 @@ class SlabGenerator:
             )
 
             # Ensure lattice a and b is consistent between the OUC and the slab
+            # Gamma remains unchecked, so the surface could diverge even with no allow_smaller_than_ouc
             if self.allow_smaller_than_ouc or (
                 np.isclose(prim_slab_l.a, prim_ouc.lattice.a) and np.isclose(prim_slab_l.b, prim_ouc.lattice.b)
             ):
--- a/src/pymatgen/core/surface.py
+++ b/src/pymatgen/core/surface.py
@@ -152,7 +152,6 @@ class Slab(Structure):
                 lattice.gamma,
             )
 
-            oriented_unit_cell = copy.deepcopy(oriented_unit_cell)
             ouc_lattice = oriented_unit_cell.lattice
             ouc_lattice = Lattice.from_parameters(
                 ouc_lattice.a,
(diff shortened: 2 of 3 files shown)


shyuep:

Solid change and CI is green across all platforms (lint + tests on Linux/macOS/Windows). Notes based on a static read of the diff only — no code was executed:

- **`allow_smaller_than_ouc`**: cleanly documented and threaded through both `SlabGenerator.__init__` and `generate_all_slabs`. Good catch moving the `energy *= prim_slab.volume / struct.volume` scaling inside the consistency branch — previously energy was rescaled even when `prim_slab` was rejected, which was a latent bug.
- **Recomputed `scale_factor`** (`np.linalg.solve(parent.matrix.T, struct.matrix.T).T`): this is a public-facing behavior change — the returned `scale_factor` becomes near-/non-integer and `c` now spans the full slab rather than the non-vacuum part. Please flag this in the changelog and confirm no internal consumers (e.g. adsorption/Wulff workflows) depend on the old integer form.
- `self.oriented_unit_cell = type(oriented_unit_cell)(...)` preserving the input type instead of hard-coding `Structure` is a nice touch; since a fresh object is constructed from species/coords, dropping the `copy.deepcopy` looks safe.
- **Test coverage**: `test_computed_scale_factor` exercises the scale-factor path, but I don't see a test covering `allow_smaller_than_ouc=True`. Worth adding one to lock in the reduced-cell / diverging-lattice behavior.

Nothing blocking from a static review.

Sanftperlig:
I couldn't find any usage of the `scale_factor` in pymatgen itself that requires the scale_factor emitted by the `SlabGenerator` to be integer (and not almost-integer) and all tests pass, so I assume it is fine? Some external users will certainly need `np.rint(scale_factor)` (or equivalent functions), but I would think if you do need integer scale factors you hopefully already round (and don't use something like `scale_factor.astype(np.int64)`)? The only non-trivial change would be to get the old `scale_factor` for truly primitive slabs back, but I don't see where that factor would be needed (why not use `primitive=False` then)?

I am still open to reducing the c in the `scale_factor` to only the actual structure part, if that is preferred.

On the test coverage: This is the last test of this PR:
```py
# We actually want [010] x [42-1], which have a 90° gamma angle:
# Due to the rhombohedral centering, the [42-1] vector can be shortened to a third of the full [42-1]
# The primitive slab will diverge in the slab and OUC lattice because .get_primitive_structure struggles
# with this primitivization (requires a "good" shift like 0.125, not a "bad" one like 0)
# Thus allow_smaller_than_ouc must be enabled
slabgen.primitive = True
slabgen.allow_smaller_than_ouc = True
p_slab = slabgen.get_slab(0.125)
# We get [010] x [-4-21]/3 from this configuration (note that c also changes to be nearly orthogonal)
assert np.allclose(p_slab.scale_factor, [[0, 1, 0], [-4 / 3, -2 / 3, 1 / 3], [4, 2, 1]])
# Check expected lattice params a/b/gamma again
expected = (5.0, 8.08977406634106, 90.0)
lat = p_slab.lattice
assert np.allclose((lat.a, lat.b, lat.gamma), expected)
```
It uses `allow_smaller_than_ouc = True`. I intentionally did not add a test to check that the OUC or the slab in bad-shift scenarios is poorly primitivized (that is unwanted behaviour after all - ideally the OUC and slab would match and both find the smaller case).

PR #106, opened 2026-07-27
Change typing to correct inaccuracy in #101

Sanftperlig:
In #101, I left `cell = "conventional"` in the docstring, even though the default is `"input"`.
This PR fixes that inaccuracy and also cleans up the default values of the overloads, as they do not require defaults.

## Checklist


shyuep:
Thanks.

PR #108, opened 2026-07-28
`__getitem__` typing for `SiteCollection` and subclasses (e. g. `Structure`)

Sanftperlig:
This PR adds a few small fixes, of which the two noteworthy ones are the following:
1. `SiteCollection.__getitem__` now has typing via a `TypeVar`, which `(I)Structure`s set to `PeriodicSite` and `Molecule`s set to `Site`. Thus, Molecules now correctly type that they use `Site` and note `PeriodicSite`. The immutable types now return their SiteType or a `Sequence` of their site type (see a)), while the mutable types return an explicit `list` instead of the `Sequence`.
2. `Slab`s now preserve the reconstruction and reorient_lattice properties through copies and similar functions. One case where this matters is if an (intentionally) non-reoriented Slab is dumped and then reloaded, currently it will be forcibly reoriented.

a) This typing should be mostly transparent to users, perhaps with some typechecker warnings as pymatgen is now explicit about what the returns are. I believe it is not possible to more strictly type the immutable collections (which return a tuple as a sequence), as they are superclasses of the mutable versions. Also, I think the single item overload override (in `Structure`/`Molecule`) is not necessary, but it should reduce the confusion of what actually happens with the types.

### DIFF
--- a/src/pymatgen/core/local_env.py
+++ b/src/pymatgen/core/local_env.py
@@ -223,8 +223,7 @@ def _handle_disorder(structure: Structure, on_disorder: on_disorder_options):
     # As a workaround, we create a new structure with majority species on each site.
     structure = structure.copy()  # make a copy so we don't mutate the original structure
     for idx, site in enumerate(structure):
-        max_specie = max(site.species, key=site.species.get)
-        max_val = site.species[max_specie]
+        max_specie, max_val = max(site.species.items(), key=lambda item: item[1])
         if max_val <= 0.5:
             if on_disorder == "take_majority_strict":
                 raise ValueError(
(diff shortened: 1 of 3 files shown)


shyuep:

Clean, well-scoped typing improvement. Making `SiteCollection` `Generic[SiteType]` and adding `@overload` signatures for `__getitem__` on `Structure`/`Molecule` is the right way to get precise element types without the old `type: ignore`. The mutable subclasses correctly narrow `sites`/`__getitem__` to concrete `list[...]` return types.

A few incidental fixes are bundled in and worth calling out since they change runtime behavior (not just types):
- `numerical_tol` is now actually forwarded in `_get_neighbor_list_py` instead of the hard-coded `1e-8` — a real bug fix; a short regression test would be good.
- `self._charge is not None` guard in `__mul__` correctly handles `charge == 0`.
- `max(site.species.items(), ...)` in `_handle_disorder` is equivalent but cleaner.

All CI green. LGTM once the neighbor-list `numerical_tol` fix has minimal test coverage.

shyuep:
Great work. Thanks.

PR #109, opened 2026-07-30
Remove ignored/deprecated ENAUG when PREC is set to Accurate

### LINKED ISSUE #4301: ENAUG in `MPScanRelaxSet.yaml`
### Python version

N/A

### Pymatgen version

2025.2.18

### Operating system version

_No response_

### Current behavior

@esoteric-ephemera, @rkingsbury:

In [MPSCANRelaxSet.yaml](https://github.com/materialsproject/pymatgen/blob/master/src/pymatgen/io/vasp/MPSCANRelaxSet.yaml), `ENAUG` is set to 1360. However, ENAUG has no effect if PREC is set to Accurate (as is the case here). It is also deprecated. See [ENAUG in the VASP manual](https://www.vasp.at/wiki/index.php/ENAUG). 

This parameter should probably be removed.

### Expected Behavior

See above.

### Minimal example

```Python

```

_No response_

DanielYang59:
close https://github.com/materialsproject/pymatgen/issues/4301

https://vasp.at/wiki/index.php/ENAUG
> Deprecated: ENAUG is considered as deprecated and should not be used anymore.

> Warning: Setting ENAUG has an effect only if [PREC](https://vasp.at/wiki/PREC) is set to one of the old settings (Low, Medium or High), otherwise it is ignored.


src/pymatgen/io/vasp/sets.py
```diff
@@ -2470,7 +2470,6 @@ def incar_updates(self) -> dict[str, Any]:
         updates = {
             "EDIFF": 1e-5,
             "EDIFFG": -0.05,
-            "ENAUG": 4000,
```
DanielYang59, line 2473: I think `MVLSlabSet` is also using `PREC = Accurate` via `MPRelaxSet`:

https://github.com/materialsproject/pymatgen-core/blob/39c044caa68d18af47318011464845a17707484a/src/pymatgen/io/vasp/sets.py#L2443

https://github.com/materialsproject/pymatgen-core/blob/39c044caa68d18af47318011464845a17707484a/src/pymatgen/io/vasp/MPRelaxSet.yaml#L79

tests/io/vasp/test_sets.py
DanielYang59, line 102: @mkhorton @janosh

shyuep:
Thanks. ENAUG is actually quite problematic. The 1360 value made calculations very expensive.

PR #110, opened 2026-07-30
Improve R-centering in `get_conventional_to_primitive_transformation_matrix` and `get_primitive_standard_structure`

Sanftperlig:
With this PR, `get_conventional_to_primitive_transformation_matrix` and `get_primitive_standard_structure` are improved:
1. `get_conventional_to_primitive_transformation_matrix` now always returns a positive determinant matrix, meaning `trans_matrix @ lattice_matrix` now never reflects the lattice. This is only a change for `R`-centering, as all other matrices already followed this rule (see a)).
2. `get_primitive_standard_structure` now doesn't re-check if coordinates are periodic images of each other in rhombohedral cells after rotating the lattice, as a lattice rotation should never lead to that being the case. This (combined with in-place editing of the lattice of the new sites) also improves the performance of this branch.
3. `international_monoclinic` is now explicitly deprecated - I assume this is preferred to leaving an unused argument forever.
4. `get_conventional_to_primitive_transformation_matrix` now has tests for the exact expected matrices.

a) In the Setyawan/Curtarolo paper I can only find the transformations for the non-R cases (I have added them as comments). Those all match and have positive determinants. Only the R case had a negative determinant (ever since the matrix was added in https://github.com/materialsproject/pymatgen-core/commit/33d5fa3f27b4d6af0eddbe2d632736ca5123e419), which would affect chiral structures, as the resulting matrix from `trans_matrix @ lattice_matrix` would be left-handed. Note that `get_primitive_standard_structure` (until https://github.com/materialsproject/pymatgen-core/commit/b111eb42c432bf1c0483055a0300d0e5d4aaa0b4 and since #88) "fixed" this for achiral structures by recreating the lattice as a right-handed lattice. Therefore the actual impact is rather small, assuming `get_primitive_standard_structure` is used significantly more than `get_conventional_to_primitive_transformation_matrix`.
The new matrix switches the first and second row to have a +1/3 determinant instead of -1/3 and it matches [spglib](https://github.com/spglib/spglib/blob/a6b561fb60cdd021a1ac90852c3c2ae14405b2c8/src/spacegroup.c#L347-L349) (transposed):
```c
static double R_mat[3][3] = {{2. / 3, -1. / 3, -1. / 3},
                             {1. / 3, 1. / 3, -2. / 3},
                             {1. / 3, 1. / 3, 1. / 3}};
```

## Checklist

### DIFF
--- a/src/pymatgen/symmetry/analyzer.py
+++ b/src/pymatgen/symmetry/analyzer.py
@@ -549,20 +549,25 @@ class SpacegroupAnalyzer:
         if space_group.startswith("P"):
             return np.eye(3, dtype=np.float64)
 
+        # Hexagonal to rhombohedral - for S/C orientation see get_primitive_standard_structure
         if space_group.startswith("R"):
             return np.array([[-1, 1, 1], [2, 1, 1], [-1, -2, 1]], dtype=np.float64) / 3
 
+        # Setyawan/Curtarolo A.3/A.5/A.8
         if space_group.startswith("I"):
             return np.array([[-1, 1, 1], [1, -1, 1], [1, 1, -1]], dtype=np.float64) / 2
 
+        # S/C A.2/A.7
         if space_group.startswith("F"):
             return np.array([[0, 1, 1], [1, 0, 1], [1, 1, 0]], dtype=np.float64) / 2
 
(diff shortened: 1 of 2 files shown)


shyuep:
Thanks for this fix.

PR #111, opened 2026-07-30
Make `SlabGenerator.gen_possible_terminations` a public function

Sanftperlig:
Currently, `gen_possible_terminations` is an inner function of `SlabGenerator.get_slabs`. With this change it becomes a public function and it can thus be accessed by external users.
This is useful, as it both allows users to see the maximum slab count `get_slabs` can return for this case (reduced by bond breakage and removal of symmetric slabs) and to access a decent termination (compare #105), which will primitivize properly with `Structure.get_primitive_structure`.

The code of the function itself stays unchanged, I only removed some unneeded typing (mypy can infer that).

Side note: Is it intentional that `repair_broken_bonds` is only called if the amount of broken bonds exceeds `max_broken_bonds` (and then allows those slabs to be used)? Personally the docstring (without seeing the code) would make me believe that `max_broken_bonds` is always the maximum number of broken bonds (meaning the amount of returned slabs is dependent on `max_broken_bonds`, but not `repair_broken_bonds`). `repair_broken_bonds` would then repair those slabs.

Also, @DanielYang59 I am not sure if I understand this comment: https://github.com/materialsproject/pymatgen-core/blob/39c044caa68d18af47318011464845a17707484a/src/pymatgen/core/surface.py#L1295-L1303 Doesn't this line account for PBC in z via `abs(z_dist - round(z_dist))`?
https://github.com/materialsproject/pymatgen-core/blob/39c044caa68d18af47318011464845a17707484a/src/pymatgen/core/surface.py#L1301

### DIFF
--- a/src/pymatgen/core/surface.py
+++ b/src/pymatgen/core/surface.py
@@ -1236,6 +1236,61 @@ class SlabGenerator:
             energy=energy,
         )
 
+    def gen_possible_terminations(self, ftol: float) -> list[float]:
+        """Generate possible terminations by clustering z coordinates.
+
+        Args:
+            ftol (float): Threshold for fcluster to check if
+                two atoms are on the same plane.
+
+        Returns:
+            termination_shifts (list[float]): Shifts for terminations that
+        """
+        frac_coords = self.oriented_unit_cell.frac_coords
+        n_atoms = len(frac_coords)
+
+        # Skip clustering when there is only one atom
+        if n_atoms < 2:
+            # Put the atom to the center
+            termination = frac_coords[0][2] + 0.5
+            return [termination - math.floor(termination)]
+
+        # Compute a Cartesian z-coordinate distance matrix
+        # TODO (@DanielYang59): account for periodic boundary condition
+        dist_matrix = np.zeros((n_atoms, n_atoms), dtype=np.float64)
+        for i, j in itertools.combinations(range(n_atoms), 2):
+            if i != j:
+                z_dist = frac_coords[i][2] - frac_coords[j][2]
+                z_dist = abs(z_dist - round(z_dist)) * self._proj_height
+                dist_matrix[i, j] = z_dist
+                dist_matrix[j, i] = z_dist
+
+        # Cluster the sites by z coordinates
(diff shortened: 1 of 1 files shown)


DanielYang59:
>  I only removed some unneeded typing (mypy can infer that).

I think some of the typing might be intended for human readers...

> Doesn't this line account for PBC in z via abs(z_dist - round(z_dist))?

Yes I think you are right, I just read this again and PBC should have been accounted for already, this comment should be removed. (should have misread in https://github.com/materialsproject/pymatgen/pull/3691#discussion_r1554557421)

src/pymatgen/core/surface.py
```diff
@@ -1236,6 +1236,61 @@ def get_slab(
...
+        if n_atoms < 2:
+            # Put the atom to the center
+            termination = frac_coords[0][2] + 0.5
+            return [termination - math.floor(termination)]
+
+        # Compute a Cartesian z-coordinate distance matrix
+        # TODO (@DanielYang59): account for periodic boundary condition
+        dist_matrix = np.zeros((n_atoms, n_atoms), dtype=np.float64)
+        for i, j in itertools.combinations(range(n_atoms), 2):
+            if i != j:
```
DanielYang59: This check (`if i != j:`) should be unnecessary, `itertools.combinations(range(n_atoms), 2)` would not produce `i == j`


```python
import itertools

for i, j in itertools.combinations(range(5), 2):
    print(i, j)
```

```
0 1
0 2
0 3
0 4
1 2
1 3
1 4
2 3
2 4
3 4
```

Sanftperlig:
I applied the comments by @DanielYang59 and I reformatted the function a bit, with one functional change:
The z coordinate for clusters is now a mean instead of the last coordinate of that cluster. I think that is better because it avoids the randomness of the site order potentially affecting the slabs (given we use single linkage, a single cluster can be quite large in z and therefore this could matter).
It passes all tests unchanged, so I assume it is fine - if you think it's unnecessary (given as the shift delta to before will likely be on the magnitude of `ftol`) or if I missed something that makes using the atom coordinate relevant, I will change it back.

PR #112, opened 2026-07-30
fix latexify_spacegroup

minhsueh:
### Problem:
`latexify_spacegroup` used `_(\d+)` to greedily capture one or more digits following an underscore. In Hermann–Mauguin notation, subscript digits (screw axes) are always a single digit — the rotation order is restricted to 1, 2, 3, 4, or 6, and the screw subscript is always less than that order (e.g. 6₁–6₅, 4₁–4₃, 3₁–3₂, 2₁). When multiple axes are concatenated with no separator, the greedy match incorrectly consumes the next axis's rotation-order digit into the current subscript.

### Fix:
Change _(\d+) → _(\d) everywhere so exactly one digit is captured per subscript.

### Example — space group 19:
```
Input:    P2_12_12_1
Original: P2_12_12₁
Fix:      P2₁2₁2₁
```

### Tests:
Added two test cases: `P2_12_12_1` -> `P2₁2₁2₁` (space group 19) and `P3_221` -> `P3₂21` (space group 154)

## Checklist

### DIFF
--- a/src/pymatgen/util/string.py
+++ b/src/pymatgen/util/string.py
@@ -248,7 +248,7 @@ def latexify_spacegroup(spacegroup_symbol: str) -> str:
     Returns:
         str: A latex formatted spacegroup with proper subscripts and overlines.
     """
-    sym = re.sub(r"_(\d+)", r"$_{\1}$", spacegroup_symbol)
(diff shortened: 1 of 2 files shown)


shyuep:
Thanks for the fix.

PR #113, opened 2026-07-31
Add the `get_symmetrically_distinct_miller_indices` cell parameter to `generate_all_slabs`

Sanftperlig:
Adds the `cell` parameter (#101) from `get_symmetrically_distinct_miller_indices` to `generate_all_slabs` so it can be used.

## Checklist


shyuep:
Thanks!

PR #114, opened 2026-07-31
Fix incorrect determinant calculation in `hkl_transformation`, add `uvw_transformation`

Sanftperlig:
This PR greatly changes `hkl_transformation`:
Currently, `hkl_transformation` calculates the denominator of the transformation matrix elements via
```py
[int(1 / i) for i in itertools.chain(*transf) if i != 0]
```
This works if the fraction is `1/n`, but for other cases (e. g. the `2/3` in R-decentering) the result is wrong.
Also, the previous hkl convention was
```py
sum(idx < 0 for idx in transf_hkl) <= 1
```
Meaning no more than one index can be negative. This is problematic for indices like (-1, 0, 0), where (1, 0, 0) is preferred. This has been changed to the same convention as in #101: First index positive. **This may affect the returned hkl in some cases, where both are reasonable, e. g.  now (1, -1, -2) was (-1, 1, 2) before.**

Also, `uvw_transformation` has been added, using the same code (but avoiding the index inversion) with the transposed inverted matrix instead of the matrix itself.

## Checklist

### DIFF
--- a/src/pymatgen/core/surface.py
+++ b/src/pymatgen/core/surface.py
@@ -2277,25 +2277,27 @@ def _is_in_miller_family(
 
 def hkl_transformation(
     transf: NDArray,
-    miller_index: tuple[int, ...] | NDArray,
+    miller_index: tuple[int, int, int] | NDArray,
 ) -> tuple[int, int, int]:
     """Transform the Miller index from setting A to B with a transformation matrix.
 
     Args:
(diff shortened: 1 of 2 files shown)


Sanftperlig:
Side note: I am aware `np.linalg.solve` would be more performant for the `uvw` case, but it doesn't allow reusing the core function and `uvw_transformation` ist only meant to be a small utility function.

PR #117, opened 2026-08-05
Fix parsing of vasprun.xml from MD Simulations Using MLFF

### LINKED ISSUE #4200: Parsing Failure of vasprun.xml from MD Simulations Using MLFF
### Python version

Python 3.10.14

### Pymatgen version

Version: 2024.11.13

### Operating system version

Red Hat Enterprise Linux Ver 8.6

### Current behavior

When trying to parse the `vasprun.xml` file generated from calculations using MLFF with the `Vasprun` class, an `IndexError: list index out of range` is raised at line 625 of `pymatgen.io.vasp.outputs.py` when accessing `self.ionic_steps[-1]["electronic_steps"]`. This seems to occur because, in MLFF calculations, data is stored in `md_data` instead of `ionic_steps`, leaving `ionic_steps` empty. Additionally, when `ionic_step_offset` or `ionic_step_skip` is specified during the creation of a Vasprun object, parsing fails because the `vasprun.xml` generated by MLFF calculations does not contain the `<calculation>` tag, which prevents parsing from working as expected.

### Expected Behavior

Parsing the `vasprun.xml` file generated from MD calculations using MLFF should not result in an error. While I am not entirely familiar with the code, changing the condition at line 348 in `outputs.py` from

```python
if self.incar.get("ALGO") not in {"Chi", "Bse"} and not self.converged and self.parameters.get("IBRION") != 0:
```
to

```python
if self.incar.get("ALGO") not in {"Chi", "Bse"} and self.parameters.get("IBRION") != 0 and not self.converged:
```
allowed me to parse the file successfully for now. However, errors still occur when `ionic_step_offset` or `ionic_step_skip` are used, and it is possible that there are other parts of the code where `ionic_steps` are referenced when `md_data` should be used instead.

### Minimal example

```Python
from pymatgen.io.vasp.outputs import Vasprun
vasprun_xml = Vasprun("vasprun.xml")
```

I am attaching a sample vasprun.xml file created for the purpose of reproducing this issue.
[vasprun.zip](https://github.com/user-attachments/files/17923967/vasprun.zip)

DanielYang59:
fix https://github.com/materialsproject/pymatgen/issues/4200


DanielYang59:
@yutoml apologize, I added the original vasprun.xml you included as test file but I forgot to ask permission for this... I hope this is fine with you or otherwise we're happy to remove that

PR #121, opened 2026-08-11
Accelerate ``FermiDos.get_e_h_concs`` (and thus ``get_doping``)

kavanase:
## Summary

This small change adds a major acceleration for `FermiDos.get_e_h_concs` (and hence `FermiDos.get_doping` and `FermiDos.get_fermi`); making it 3–10+× faster with identical results (to within ~float noise, ~1e-12 _relative_ error). 

Two main changes (in `dos.py`):

1. Precomputed trapezoid weights in `FermiDos.__init__`: because we have a fixed energy grid, `trapezoid(y, x)` is just a dot product `w @ y`, so this removes per-call `scipy` dispatch overhead and the weights are folded with `tdos` once (rather than every call). 
2. Fermi-Dirac truncation: the Fermi-Dirac distribution decays as `exp(-|E−E_F|/kT)` away from the Fermi level, so beyond ~40 k_BT the contribution is negligible (~<1e-17, relative) and thus can be truncated, greatly reducing the integration range.

Passes built-in DOS tests, and tested extensively indirectly via the `doped` defect/carrier concentration analyses for a range of system DOSs and Fermi Levels. Speedup of ~3-10+x.

## Checklist

- [ ] Google format doc strings added. Check with `ruff`.

### DIFF
--- a/src/pymatgen/electronic_structure/dos.py
+++ b/src/pymatgen/electronic_structure/dos.py
@@ -465,6 +465,12 @@ class FermiDos(Dos, MSONable):
             self.energies[: self.idx_mid_gap] -= (bandgap - (ecbm - evbm)) / 2.0
             self.energies[self.idx_mid_gap :] += (bandgap - (ecbm - evbm)) / 2.0
 
+        # Precompute trapezoid integration weights folded with the DOS, so ``get_e_h_concs`` reduces to a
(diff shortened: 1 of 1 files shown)


shyuep:
Nice. Thanks!

PR #123, opened 2026-08-12
Note Enumlib.jl as a drop-in enum.x successor for EnumlibAdaptor

glwhart:
## Summary

I'm Gus Hart — a co-author of the Fortran `enumlib` (`enum.x` / `makeStr.py`) that `EnumlibAdaptor` wraps. I've rewritten the enumeration engine from scratch in Julia ([**Enumlib.jl**](https://github.com/glwhart/Enumlib.jl)) and verified it as a **drop-in replacement for `enum.x`** — concretely, **this module's own `test_enumlib_caller.py` passes unmodified against it, green in CI:**

**→ https://github.com/glwhart/pymatgen-core/actions/runs/30926741986** (4/4 passed)

**This PR is intentionally tiny:** it adds a short note to the `enumlib_caller` module docstring pointing users to Enumlib.jl as an alternative provider of `enum.x`. **No runtime or CI changes.** I'd mainly like your read on whether/how you'd want to surface this to users — happy to expand it, trim it, or take a different route entirely.

## Why it might interest your users — over the Fortran `enum.x`

- **Catchable exceptions with messages** instead of silent nonzero exits / runaway subprocesses — the class of friction behind issues like #4185 (timeout not honored through the subprocess) and #1063.
- **A pre-flight cost + memory estimate** (overridable) so a combinatorially explosive request is refused/warned up front rather than churning or OOM-ing.
- **Count-without-enumerating** (Pólya) — report how many symmetry-distinct structures a request yields *before* generating them.
- Fixed-concentration / concentration-range enumeration, a recursive-stabilizer algorithm, automatic algorithm selection.
- Modern, tested (~1,900 tests), documented, semver'd, actively developed.

## Why "it's Julia" needn't reach your users

The engine ships as a **standalone `enum.x` with the Julia runtime baked into the binary** (PackageCompiler) — no Julia install, no `juliacall`, nothing Julia-visible — distributable via conda-forge exactly like today's enumlib. Only `enum.x` is swapped; `makeStr.py` is reused unchanged.

## Evidence

- **Your suite, green** against the Julia `enum.x` (link above): LiFePO₄ → 86, Cu₇Te₅ → 197, garnet → 7/20, timeout handling.
- **End-to-end via `EnumlibAdaptor`, unchanged:** Fortran vs. Julia `enum.x` on the same disordered structure → identical `Structure`s (`StructureMatcher`).
- **Count parity with the Fortran `enum.x`** to **~166M** structures (fcc/bcc/sc, hcp, diamond) and **~50M** for a multilattice *ternary* case, correct across Bravais classes.

## Honest notes

- **Footprint:** the self-contained binary is ~650 MB (bundled Julia runtime), ~200–250 MB compressed. An optional in-process (`juliacall`) path exists for a lean install — opt-in only, never a core dependency.
- **Magnetic enumeration** (the `MagneticStructureEnumerator` path) needs one more piece, targeted for the next release; everything else is current.
- **One deliberate difference from Fortran, in a narrow corner:** when two symmetry-equivalent sublattices are assigned *different* species sets, Enumlib.jl returns the mathematically correct count for the full space group; the Fortran convention conflates them and returns a different number. Your suite doesn't hit this, but I'd document it in a short migration note.

## How I'd like to proceed

This PR is a conversation-opener. Merge the note as-is, tell me to trim/drop it, or point me toward what you'd actually find useful — a compatibility CI job, conda-forge packaging, a docs section, or the optional `juliacall` path. I'd especially value your read on the **packaging** choice given your users' install constraints. Glad to demo or discuss first.

### DIFF
--- a/src/pymatgen/command_line/enumlib_caller.py
+++ b/src/pymatgen/command_line/enumlib_caller.py
@@ -6,6 +6,13 @@ makestr.x available in the path. Please download the library at
 https://github.com/msg-byu/enumlib and follow the instructions in the README to
 compile these two executables accordingly.
 
+Alternatively, ``enum.x`` can be provided by Enumlib.jl
+(https://github.com/glwhart/Enumlib.jl), a from-scratch Julia reimplementation of
+enumlib by one of its authors. It is a drop-in replacement: the
+``struct_enum.in`` / ``struct_enum.out`` file contract is unchanged and
+``makeStr.py`` is reused, so only ``enum.x`` is swapped. This module's own test
+suite passes against it.
+
 If you use this module, please cite:
 
     - Gus L. W. Hart and Rodney W. Forcade, "Algorithm for generating derivative
--- a/src/pymatgen/command_line/enumlib_caller.py
+++ b/src/pymatgen/command_line/enumlib_caller.py
@@ -33,11 +33,13 @@ If you use this module, please cite:
 from __future__ import annotations
 
 import fractions
+import functools
 import itertools
 import logging
 import math
 import re
 import subprocess
+import warnings
 from glob import glob
 from shutil import which
 from typing import TYPE_CHECKING
@@ -61,6 +63,61 @@ logger = logging.getLogger(__name__)
 ENUM_CMD = which("enum.x") or which("multienum.x")
 # Prefer makestr.x at present
 MAKESTR_CMD = which("makestr.x") or which("makeStr.x") or which("makeStr.py")
+
+
+@functools.cache
+def _is_enumlib_jl(enum_cmd: str | None) -> bool:
+    """Detect whether the resolved enum.x is the Julia Enumlib.jl engine.
+
+    Enumlib.jl responds to ``enum.x --version`` by printing a line containing
+    the token ``Enumlib.jl`` (e.g. ``enum.x (Enumlib.jl) 0.3.2``). The legacy
+    Fortran enum.x does not support ``--version``. The probe is cached so the
+    subprocess is spawned at most once per process per command path, and it
+    never raises: any failure is treated as "not Enumlib.jl".
+
+    Args:
+        enum_cmd (str | None): Resolved path to the enum.x executable.
+
+    Returns:
+        bool: True iff the executable identifies itself as Enumlib.jl.
+    """
+    if not enum_cmd:
+        return False
+    try:
+        proc = subprocess.run(
+            [enum_cmd, "--version"],
+            capture_output=True,
+            text=True,
+            timeout=10,
+            check=False,
+        )
+        return "Enumlib.jl" in (proc.stdout or "") + (proc.stderr or "")
+    except Exception:
+        # Never let the version probe crash enumeration: any failure
+        # (FileNotFoundError, subprocess.TimeoutExpired, OSError, ...) simply
+        # means we cannot confirm the Julia engine.
+        return False
+
+
+def _warn_if_not_enumlib_jl(enum_cmd: str | None) -> None:
+    """Advise switching to Enumlib.jl when the legacy Fortran enum.x is used.
+
+    Purely advisory: this never raises or changes enumeration behavior. The
+    underlying probe is cached, so the subprocess runs at most once per command
+    path; how often the warning itself is shown is up to the active warning
+    filters (Python's default filter shows it once per call site).
+
+    Args:
+        enum_cmd (str | None): Resolved path to the enum.x executable.
+    """
+    if enum_cmd and not _is_enumlib_jl(enum_cmd):
+        warnings.warn(
+            "You appear to be using the legacy Fortran 'enum.x'. Consider installing "
+            "Enumlib.jl (https://github.com/glwhart/Enumlib.jl), a faster, "
+            "better-maintained drop-in replacement for enumlib.",
+            UserWarning,
+            stacklevel=2,
+        )
 
 
 @requires(
@@ -134,6 +191,11 @@ class EnumlibAdaptor:
         self.enum_precision_parameter = enum_precision_parameter
         self.check_ordered_symmetry = check_ordered_symmetry
         self.timeout = timeout
+
+        # Recommend the faster, better-maintained Enumlib.jl when the legacy
+        # Fortran enum.x is detected. Advisory only; the version probe behind
+        # this is cached, so repeated instantiation costs no extra subprocesses.
+        _warn_if_not_enumlib_jl(ENUM_CMD)
 
     def run(self) -> None:
         """Run the enumeration."""
--- a/tests/command_line/test_enumlib_caller.py
+++ b/tests/command_line/test_enumlib_caller.py
@@ -1,12 +1,16 @@
 from __future__ import annotations
 
 import re
+import subprocess
+import warnings
 from shutil import which
+from types import SimpleNamespace
 
 import numpy as np
 import pytest
 from pytest import approx
 
+from pymatgen.command_line import enumlib_caller
 from pymatgen.command_line.enumlib_caller import EnumError, EnumlibAdaptor
 from pymatgen.core import Element, Structure
 from pymatgen.symmetry.analyzer import SpacegroupAnalyzer
@@ -18,6 +22,48 @@ ENUM_CMD = which("enum.x") or which("multienum.x")
 MAKESTR_CMD = which("makestr.x") or which("makeStr.x") or which("makeStr.py")
 
 ENUMLIB_TEST_FILES_DIR: str = f"{TEST_FILES_DIR}/command_line/enumlib"
+
+
+class TestEnumlibJlWarning:
+    """Tests for the advisory Enumlib.jl recommendation.
+
+    These are independent of any installed enum.x: the ``--version`` probe is
+    monkeypatched to simulate the Fortran and Julia engines.
+    """
+
+    @staticmethod
+    def _patch_version_output(monkeypatch, stdout="", stderr=""):
+        """Fake ``subprocess.run`` so the version probe returns fixed output."""
+
+        def fake_run(*_args, **_kwargs):
+            return SimpleNamespace(stdout=stdout, stderr=stderr, returncode=0)
+
+        monkeypatch.setattr(enumlib_caller.subprocess, "run", fake_run)
+        # Clear the per-process probe cache so each simulation is honored.
+        enumlib_caller._is_enumlib_jl.cache_clear()
+
+    def test_warns_for_fortran_enum(self, monkeypatch):
+        # Legacy Fortran enum.x: no "Enumlib.jl" token -> warning fires.
+        self._patch_version_output(monkeypatch, stderr="Unknown option --version")
+        with pytest.warns(UserWarning, match="Enumlib.jl"):
+            enumlib_caller._warn_if_not_enumlib_jl("/fake/fortran/enum.x")
+
+    def test_no_warn_for_julia_enum(self, monkeypatch):
+        # Julia Enumlib.jl engine: token present -> no warning.
+        self._patch_version_output(monkeypatch, stdout="enum.x (Enumlib.jl) 0.3.2")
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")  # any UserWarning would fail the test
+            enumlib_caller._warn_if_not_enumlib_jl("/fake/julia/enum.x")
+        assert enumlib_caller._is_enumlib_jl("/fake/julia/enum.x") is True
+
+    def test_probe_never_raises(self, monkeypatch):
+        # A crashing probe must be swallowed and reported as "not Enumlib.jl".
+        def boom(*_args, **_kwargs):
+            raise subprocess.TimeoutExpired(cmd="enum.x", timeout=10)
+
+        monkeypatch.setattr(enumlib_caller.subprocess, "run", boom)
+        enumlib_caller._is_enumlib_jl.cache_clear()
+        assert enumlib_caller._is_enumlib_jl("/fake/hanging/enum.x") is False
 
 
 @pytest.mark.skipif(not (ENUM_CMD and MAKESTR_CMD), reason="enumlib not present.")


shyuep:
Thanks @glwhart ! This is great and all features greatly appreciated by the community! Here are some suggestions:

1. Might be good to modify enumlib_caller so that if the old enum.x is detected, an explicit (non-breaking) warning is raised to recommend installation of Enumlib.jl. Not sure if there is a simple way to detect this. Maybe just a quick file size check (< or > 300Mb?) would be sufficient?
2. I believe right now enumlib is installed via conda-forge in our CI workflows. One thing you might want to consider is to request the maintainer of the conda-forge version of enumlib to modify the build to use Enumlib.jl instead. Or is this already the case today? If so, then nothing else is needed since the CI will use the Julia version.

I can help make the recommendation to use Enumlib.jl more explicit in the pymatgen docs, including the READMEs, etc.

glwhart:
Thanks so much, @shyuep — glad it's welcome!

**1. Warn on the old `enum.x`.** Happy to add. Rather than a size heuristic (the footprint varies a lot — ~650 MB for the self-contained binary vs. a small `juliacall`-based install), I'll make it robust: Enumlib.jl's `enum.x` will answer `enum.x --version` with an identifiable string, and I'll add a small, non-breaking `warnings.warn(...)` in `enumlib_caller` that recommends Enumlib.jl when that probe is absent. I can put it in this PR or a follow-up — your preference.

**2. conda-forge, and retiring the Fortran.** Not yet — the current conda-forge `enumlib` is still the Fortran build, so your CI uses Fortran today. Candidly, I don't intend to keep maintaining the Fortran `enumlib` — Enumlib.jl supersedes it and is the better engine — so the end state I'm aiming for is conda-forge's `enumlib` delivering the Julia `enum.x`, with the Fortran retired. As the original `enumlib` author I'll drive that and work with the feedstock maintainers to stage it so nothing breaks mid-switch; once it's in place, your CI and users get the Julia engine transparently, with no change on your side. If you have a pulse on who currently maintains that feedstock, that'd help.

**3. Docs/READMEs** — yes please, I'd gladly take you up on that, and I'm happy to draft text alongside.

(FWIW the one red CI check is a transient conda-forge fetch failure at the "Create mamba environment" step on the ubuntu-3.11 cell — unrelated to this docstring change; a re-run should clear it.)

shyuep:
Yes, just include the warning in this same PR would be appreciated.

Once merged, I will draft a quick branch of the CI workflow to directly install from the Enumlib.jl CI just to make sure everything is working.

shyuep (commented):

Docs-only change: a short `enumlib_caller` docstring note pointing to Enumlib.jl as a drop-in `enum.x` provider, with the module's own tests passing against it. Low risk, and the linked CI evidence is a nice touch.

- The one failing check (`test ubuntu-latest, 3.11, lowest-direct`) is an infra flake (`fetch failed` during setup, no test errors) — re-running should clear it.
- Fine to land as-is; deeper integration (e.g. documented conda-forge install path or an `ENUM_CMD` hint) can be a follow-up discussion.

glwhart:
Thanks @shyuep — the warning is now in this PR (179fb81), so it's one PR as you asked.

**How the detection works.** Rather than the file-size heuristic, I added a `--version`/`-V` flag to Enumlib.jl's `enum.x`, which prints one line containing the token `Enumlib.jl`:

```
$ enum.x --version
enum.x (Enumlib.jl) 0.3.2
```

The legacy Fortran `enum.x` doesn't support the flag, so its absence is the signal. That seemed more robust than a size threshold, which would misfire on stripped or statically-linked Fortran builds.

**Non-breaking by construction.** The probe is `functools.cache`d per command path (at most one subprocess per process), and it swallows everything — missing file, timeout, `OSError` — treating any failure as "not Enumlib.jl". Enumeration behavior is untouched; the only effect is a `UserWarning`. The three new tests monkeypatch the probe, so they pass with neither engine installed. I also verified end-to-end against a real Julia `enum.x` (probe returns `True`, no warning) and a non-Julia executable (`False`, warning fires).

**One judgment call worth your review:** the warning fires in `EnumlibAdaptor.__init__` rather than in `run()`, so users see it at construction even if they never enumerate. Happy to move it to `run()` if you'd rather it be tied to actual execution.

**On your CI plan** — that sounds ideal, with one sequencing caveat: `--version` currently lives on an Enumlib.jl feature branch, not a tagged release. I'll land it on `main` and cut a release before you wire up that workflow branch, so there's a stable artifact to install from. I'll follow up here when it's tagged.

Still open from my last comment whenever you get a chance: do you know who maintains the conda-forge `enumlib` feedstock? That's the piece I need in order to make `conda install enumlib` deliver the Julia engine and retire the Fortran.

glwhart:
Another significant feature of the new enumlib is that the documentation is really good. I worked really hard on that. https://glwhart.github.io/Enumlib.jl/dev/

It has walkthroughs, tutorials, algorithm discussion, and all the api are documented well.

shyuep:
Thanks @glwhart . I will merge this PR. For conda-forge, I am not sure who the maintainer is. In the early days, I started a conda-forge version of enumlib but someone else have taken over. But if I am not wrong, you can simply follow their documentation to submit a PR to update the conda-forge recipe for enumlib and the maintainer will merge it (after some CI checks).

shyuep:
@glwhart I have done a pymatgen-core release and updated the pymatgen.org docs to highlight the new Julia implementation. One small question - how do people download the pre-compiled standalone binary? I was looking at the https://github.com/glwhart/Enumlib.jl and couldn't find a pre-compiled binary there.

glwhart:
Thanks @shyuep — and thanks for merging and for the docs update.

You couldn't find a pre-compiled binary because there wasn't one: the repo had tags but no GitHub Releases. That's fixed. There are now per-platform standalone builds on the [releases page](https://github.com/glwhart/Enumlib.jl/releases), currently v0.3.4:

```
https://github.com/glwhart/Enumlib.jl/releases/download/v0.3.4/enumlib-jl-0.3.4-linux-x86_64.tar.gz
https://github.com/glwhart/Enumlib.jl/releases/download/v0.3.4/enumlib-jl-0.3.4-macos-aarch64.tar.gz
https://github.com/glwhart/Enumlib.jl/releases/download/v0.3.4/enumlib-jl-0.3.4-windows-x86_64.tar.gz
```

Each ships with a `.sha256`. They're PackageCompiler apps that bundle the Julia runtime, so there's no Julia installation step — untar and put `bin/` on `PATH`:

```bash
curl -sSL https://github.com/glwhart/Enumlib.jl/releases/download/v0.3.4/enumlib-jl-0.3.4-linux-x86_64.tar.gz | tar xz
export PATH="$PWD/enumlib-jl-0.3.4-linux-x86_64/bin:$PATH"
enum.x --version      # enum.x (Enumlib.jl) 0.3.4
```

That should be all your CI workflow branch needs. Fair warning on size: bundling the runtime costs about 300–450 MB compressed per platform. It's the price of not requiring users to install Julia, and I'd rather pay it than make `enum.x` a two-step install.

Two things worth knowing beyond the binary itself:

**`--version` is now in a tagged release.** When you merged, it existed only on a feature branch, so the detection this PR added would have reported *every* installation — including Enumlib.jl ones — as the legacy Fortran. Anything from v0.3.3 onward identifies itself properly, so the warning now does what it says.

**The Julia package now covers all three enumlib executables, not just `enum.x`.** `polya.x` and `makestr.x` are both implemented. `makestr.x` reads `struct_enum.out` and writes `vasp.<n>` POSCARs exactly as before; I checked the output against `Poscar.from_str(data, index_species)` — the same call `EnumlibAdaptor` makes — so it should be a drop-in for the structure-generation half too. That matters for the conda-forge plan below: it means the Julia build can replace the Fortran package outright rather than leaving users to install the Fortran anyway for `makestr.x`.

On conda-forge — thanks, that's the answer I needed. I'll open a PR against the feedstock directly rather than trying to track down the maintainer. My plan is to have the recipe repackage these prebuilt tarballs rather than build in the feedstock: `create_app` needs the whole Julia depot and registry at build time, which doesn't fit conda-forge's reproducible-build norms. That does need conda-forge/core to sign off on repackaging a prebuilt binary, so I'll stage it under a separate name first and only fold it into `enumlib` once it's proven — no disruption to the current package while that plays out.

Last thing: you didn't say either way on the warning firing in `EnumlibAdaptor.__init__` rather than `run()`. As merged, users see it at construction even if they never enumerate. Happy to move it if you'd prefer it tied to actual execution — just say the word and I'll send a small follow-up PR.

shyuep:
@glwhart Thanks! Putting the warning in the init is fine. I doubt anyone would init an enumlib adaptor if they don't intend to run it. Also, it ensures that we don't fire the message multiple times. 

Just a note that the install is not as simple as untaring and running enum.x. The library path needs to be set.

shyuep:
@glwhart Just a few more things to report - to get this to work on MacOS, I had to do the following:

```bash
export DYLD_FALLBACK_LIBRARY_PATH=$DYLD_FALLBACK_LIBRARY_PATH:$PWD/enumlib-jl-0.3.4-macos-aarch64/lib

# The commands below mass-approve the dylib and framework files bundled.
sudo find $PWD/enumlib-jl-0.3.4-macos-aarch64 -name "*.dylib" -exec xattr -d com.apple.quarantine {} \;
sudo find $PWD/enumlib-jl-0.3.4-macos-aarch64 -name "*.framework" -exec xattr -d com.apple.quarantine {} \;
```

Thereafter, the `enum.x --version` command yields `enum.x (Enumlib.jl) nothing`.

I recommend the following:
1. Code sign the dylib and framework files so that they don't run afoul of MacOS's security restrictions.
2. The version seems to not be tagged in the released binary.

glwhart:
Thanks for testing it on a real machine — both reports were worth having, and one of them was a genuine bug.

**The version really was broken, and it's fixed in v0.3.5.** You were right: `pkgversion` has no package `Project.toml` to read inside a PackageCompiler app, and it *returns* `nothing` rather than throwing — so my `try`/`catch` never fired and `string(nothing)` printed the literal word. The version is now resolved at precompile time into a constant, where the Project.toml is present, and frozen into the app image.

The more embarrassing part is why CI didn't catch it: my build smoke test asserted only that `--version` output contained `Enumlib.jl`, and `enum.x (Enumlib.jl) nothing` satisfies that. It now asserts the exact version string, so a build that misreports its version fails on the runner instead of shipping. Worth noting the `Enumlib.jl` token itself was always correct, so the detection this PR added was never affected — only the number.

**On the library path, I can't reproduce it, and I'd like to understand what you hit.** On a clean macOS arm64 box, the published v0.3.4 tarball runs with no `DYLD_*` variable set at all, invoked by absolute path from an unrelated working directory:

```
$ env -u DYLD_FALLBACK_LIBRARY_PATH /path/to/enumlib-jl-0.3.4-macos-aarch64/bin/enum.x --version
enum.x (Enumlib.jl) nothing
```

`otool -l` confirms the launcher carries `@executable_path/../lib` and `@executable_path/../lib/julia` as RPATHs, which is what should make it relocatable. So `DYLD_FALLBACK_LIBRARY_PATH` shouldn't be needed, and I'd rather find the real cause than have people set it.

My guess is that what you saw was the Gatekeeper problem wearing a disguise: when a bundled `.dylib` is quarantined, the failure surfaces as a load error that reads like a missing library, which naturally invites a `DYLD_*` workaround. If you still have the terminal scrollback, the exact error before you set that variable would tell us. (I tried to reproduce by force-applying `com.apple.quarantine` to the whole tree and it still ran here, so your Gatekeeper policy is evidently stricter than mine.)

**Code signing: agreed, that's the right fix.** Proper Developer ID signing plus notarization is the real answer and I'll look into getting that set up — it needs an Apple developer account on my end. In the meantime two things make it much less painful than the `sudo find` incantation:

```bash
# one command, whole tree, no sudo needed — you own the files
xattr -dr com.apple.quarantine enumlib-jl-0.3.5-macos-aarch64
```

and downloading with `curl` or `wget` instead of a browser avoids the quarantine flag being set in the first place. Both are now documented in the README along with the note that no `DYLD_*`/`LD_LIBRARY_PATH` setup should be necessary.

Good to have the `__init__` question settled — I agree on both counts, particularly that it keeps the warning from firing repeatedly. Leaving it as merged.

v0.3.5 with the version fix: https://github.com/glwhart/Enumlib.jl/releases/tag/v0.3.5

shyuep:
Thanks @glwhart . Will update the docs accordingly.

glwhart:
Three follow-ups since your report, one of which affects the CI workflow you mentioned.

**The version fix is out** (v0.3.5 onward) — `enum.x --version` reports a real number instead of `nothing`.

**More important for you: the linux binaries were broken on older glibc, and are now fixed.** Everything up to v0.3.6 was built on Ubuntu 24.04 and required `GLIBC_2.38`, so it would not start on RHEL/Rocky 8–9 or on typical HPC nodes — if you had wired CI to those tarballs it would have failed outright, and not in an obvious way. From v0.3.7 the linux artifact is built inside `manylinux_2_28`; the whole bundle now has a `GLIBC_2.17` floor, checked on every build.

**The download is about two-thirds smaller.** An upstream dependency of mine declared `Plots` without ever using it, which dragged GR, Qt6, Cairo, FFMPEG and X11 into the bundled app. Removing that one unused line took linux from 437 MB to 158 MB and macOS from 297 MB to 91 MB. Current release is v0.3.8, so I'd point CI at that rather than anything earlier.

On conda-forge: I've submitted a recipe as `enumlib-jl` (conda-forge/staged-recipes#34550) rather than touching the existing `enumlib` package — all checks green, now waiting on review. It repackages the prebuilt release artifact, which needs conda-forge/core to sign off, so I don't know the timeline. Once it's proven I'll take up folding it into the `enumlib` name with that feedstock's maintainer, as you suggested.

One loose end from your side if you have a moment: you mentioned needing `DYLD_FALLBACK_LIBRARY_PATH`, and I still can't reproduce that — the binaries carry `@executable_path/../lib` as an RPATH and run here with no `DYLD_*` set. If it still happens on v0.3.8 I'd like the exact error, since I'd rather fix it than have it documented as a workaround.

PR #124, opened 2026-08-14
fix minor logic for POTCAR validation

matthewkuner:
Minor fix for the logic around POTCAR validation. Ensures that last-digit discrepancies for very large, squared quantities do not cause a POTCAR to be marked as invalid


shyuep:
Thanks!

PR #125, opened 2026-08-14
Allow Strain initialization from Voigt vectors

### LINKED ISSUE #4549: Strain does not handle Voigt notation inputs as stated.
### Python version

Python 3.10.12

### Pymatgen version

Version: 2025.10.7

### Operating system version

Ubuntu 22.04 LTS

### Current behavior

```Strain``` states it can convert from a Voigt notation input,

```Parameters:
strain_matrix (ArrayLike) – 3x3 matrix or length-6 Voigt notation vector representing the Green-Lagrange strain
```
But doing so hits an error in ```Tensor``` (via ```SquareTensor```)

```shell
Traceback (most recent call last):
  File "/home/harvey/bug.py", line 12, in <module>
    s2 = Strain(s1.voigt)
  File "/home/harvey/.local/lib/python3.10/site-packages/pymatgen/analysis/elasticity/strain.py", line 174, in __new__
    obj = super().__new__(cls, strain_matrix, vscale=vscale)
  File "/home/harvey/.local/lib/python3.10/site-packages/pymatgen/core/tensors.py", line 940, in __new__
    obj = super().__new__(cls, input_array, vscale, check_rank=2)
  File "/home/harvey/.local/lib/python3.10/site-packages/pymatgen/core/tensors.py", line 72, in __new__
    raise ValueError(f"{type(obj).__name__} input must be rank {check_rank}")
ValueError: Strain input must be rank 2
```

It looks like ```Strain``` does [no handling](https://github.com/materialsproject/pymatgen/blob/v2025.6.14/src/pymatgen/analysis/elasticity/strain.py#L156-L237), and it calls ```SquareTensor``` [that](https://github.com/materialsproject/pymatgen/blob/v2025.6.14/src/pymatgen/core/tensors.py#L920) asks for a 3x3 matrix and so fails as it should.

This could be solved with ```Tensor.from_voigt```.

### Expected Behavior

Strain should be able to handle conversion from a 6-length Voigt vector.

```Tensor.from_voigt``` will do this.

```python
from pymatgen.analysis.elasticity import Strain                                 
from pymatgen.core.tensors import Tensor                                        
                                                                                
import numpy as np                                                              
                                                                                
s = np.array([[0., 0., -0.06], [0.0, 0.0, 0.0], [-0.06, 0.0, 0.0]])             
                                                                                
s1 = Strain(s)                                                                  
print(s1)                                                                       
                                                                                
v = s1.voigt           
# Since  in Strain vscale = np.ones((6,)); vscale[3:] *= 2                                                    
v[3:]/=2.0                                                    
                                                                                
s2 = Strain(Tensor.from_voigt(v)) # Works, s2 == s1                                     
print(s2)
```

### Minimal example

```Python
from pymatgen.analysis.elasticity import Strain
import numpy as np

s = np.array([[0., 0., -0.06], [0.0, 0.0, 0.0], [-0.06, 0.0, 0.0]])

s2 = Strain(Strain(s).voigt) # Fails
```

_No response_

Krataios14:
`Strain` is documented to accept a six-element Voigt vector, but passing one currently raises a rank error in `SquareTensor`.

This change detects that input and uses the existing `from_voigt` conversion. That also handles the factor-of-two scaling for shear components correctly.

I added a round-trip test with a nonzero shear component to cover the bug.

Fixes materialsproject/pymatgen#4549

### DIFF
--- a/src/pymatgen/core/elasticity/strain.py
+++ b/src/pymatgen/core/elasticity/strain.py
@@ -169,6 +169,9 @@ class Strain(SquareTensor):
             strain_matrix (ArrayLike): 3x3 matrix or length-6 Voigt notation vector
                 representing the Green-Lagrange strain
         """
+        if np.shape(strain_matrix) == (6,):
+            return cls.from_voigt(strain_matrix)
+
         vscale = np.ones((6,))
         vscale[3:] *= 2
         obj = super().__new__(cls, strain_matrix, vscale=vscale)
--- a/tests/core/elasticity/test_strain.py
+++ b/tests/core/elasticity/test_strain.py
@@ -93,6 +93,7 @@ class TestStrain(MatSciTest):
     def test_new(self):
         test_strain = Strain([[0, 0.01, 0], [0.01, 0.0002, 0], [0, 0, 0]])
         assert_allclose(test_strain, test_strain.get_deformation_matrix().green_lagrange_strain)
+        assert_allclose(Strain(test_strain.voigt), test_strain)
         with pytest.raises(
             ValueError,
             match="Strain must be initialized with a symmetric array or a Voigt-notation vector",


shyuep (commented):
*Automated PR review generated by Claude on behalf of @shyuep. Static analysis of the diff only; no code was run locally.*

Correct and minimal fix for materialsproject/pymatgen#4549 — the docstring already promised length-6 Voigt input, and delegating to `from_voigt` gets the factor-of-two shear scaling right. No recursion risk, since `from_voigt` re-enters `__new__` with a 3x3 array. Two suggestions:

1. **Strengthen the test.** The round-trip `Strain(test_strain.voigt)` passes even if the shear convention were wrong in a self-consistent way. Add an explicit check, e.g. `Strain([0, 0, 0, 0.02, 0, 0])[1, 2] == approx(0.01)`, to pin the factor of two.

2. **Minor:** consider `np.shape(strain_matrix) == (6,)` → also accepting `(6, 1)`/`(1, 6)`? Probably not worth it, but if you decline, the docstring should say a flat length-6 vector specifically.

CI has not run on this branch — needs a workflow approval before merge.

shyuep:
Thanks.

PR #127, opened 2026-08-14
Fix periodic interpolation of volumetric data

Krataios14:
## Summary

`VolumetricData` currently places an N-point grid on `linspace(0, 1, N)`, but volumetric samples are located at `i / N`; the endpoint belongs to the next periodic image. This shifts interpolation coordinates and prevents interpolation across the unit-cell boundary.

This PR performs trilinear interpolation directly in grid-index space. Fractional coordinates are wrapped into the unit cell, including the floating-point edge case where a tiny negative coordinate wraps to exactly `1.0`, and upper neighbors wrap back to index zero. Working from the original array also avoids making a padded copy of large volumetric grids. `linear_slice` now evaluates the complete point array in one vectorized interpolation call.

## User-visible changes

- `value_at` and `linear_slice` return corrected numerical values. For example, the ELFCAR reference at `(0.4, 0.5, 0.6)` changes from `0.0918471` to `0.1167966`.
- Coordinates outside the unit cell now wrap periodically instead of raising a bounds error.
- The call form `volumetric_data.interpolator(points)` remains available, but `interpolator` is now a method rather than a `RegularGridInterpolator` instance. Code using `.interpolator.grid`, `.interpolator.values`, or the previous `method` and `fill_value` call options will need to be updated.

These compatibility and numerical changes are recorded in `CHANGES.md`.

## Testing

- Added regression tests for exact grid values, all three periodic boundaries, integer lattice translations, tiny negative coordinates, updated data, invalid coordinate shapes, singleton dimensions, and vectorized linear slices.
- Updated the ELFCAR interpolation reference value and verified it directly from neighboring grid samples.
- Ran `tests/io/test_common.py`, `TestElfcar`, and `TestChgcar`: 17 passed and one optional HDF5 test skipped.
- Ran Ruff, formatting checks, and Pyright.

Addresses materialsproject/pymatgen#3787.


shyuep (commented):
*Automated PR review generated by Claude on behalf of @shyuep. Static analysis of the diff only; no code was run locally.*

The core diagnosis is correct — volumetric samples sit at `i/n`, not `linspace(0, 1, n)`, so the old grid was shifted and could not interpolate across the cell boundary. Nice that the trilinear kernel avoids padding a copy of large grids. Points to address:

1. **Possible `IndexError` from `np.mod`.** For tiny negative coordinates `np.mod(-1e-18, 1.0)` returns exactly `1.0` in float64, so `grid_coords == n` and `lower == n` → out-of-bounds index. Add `lower %= grid_shape` (weights still work out since `upper_weights` is then ~0).

2. **`interpolator` changes from attribute to method — breaking change.** Callers doing `vd.interpolator(pts)` are fine, but anything touching `.interpolator.values`/`.grid` or passing `method=`/`fill_value=` to the `RegularGridInterpolator` breaks, and `bounds_error=True` is silently replaced by periodic wrapping. Please call this out in the PR description so it makes the release notes.

3. **Numerical results change for users.** The `Elfcar` reference in `test_outputs.py` moves 0.0918 → 0.1168. Correct, but a user-visible change in `value_at`/`linear_slice` output that deserves an explicit changelog entry.

4. **Minor:** `linear_slice` still loops `value_at` per point (`common.py:312`); `interpolator` is now vectorized, so a single call over the whole `(n, 3)` array would be substantially faster.

Otherwise `itertools` is imported, the singleton-dimension case is handled and tested, and the regression tests cover the boundary/translation invariants well. CI has not run on this branch — needs a workflow approval before merge.

Krataios14:
Addressed all four review points in 0d50da8: lower indices now wrap after weight calculation, the tiny-negative case has a regression test, linear_slice uses one vectorized interpolation call while still returning a list, and both CHANGES.md and the PR description explicitly document the numerical and interpolator API changes. Relevant tests pass locally (17 passed, one optional HDF5 skip); Ruff, formatting, and Pyright are clean.

shyuep (commented):
*Automated PR review generated by Claude on behalf of @shyuep. Re-review after 0d50da8; static analysis of the diff only, no code executed.*

All four items from the last round are addressed, and the trilinear implementation reads correctly:

- `lower %= grid_shape` happens *after* the weights are formed, so the `np.mod(x, 1) == 1.0` case (tiny negative input) lands on index 0 with weight 1 — covered by the `value_at(-1e-18, 0, 0)` assertion.
- Singleton axes degrade correctly: `upper = (0 + 1) % 1 == 0`, so both corners hit index 0 and the weights still sum to 1.
- `linear_slice` is a single vectorized call and still returns a `list`.
- CI 5/5 green.

Two things for the record before merge:

1. **The API break is real, not just numerical.** `interpolator` goes from a `RegularGridInterpolator` attribute to a bound method, so `.grid`/`.values` and the `method=`/`fill_value=` call kwargs disappear. CHANGES.md documents it, but this both silently changes results for any downstream caller of `value_at`/`linear_slice` and hard-breaks anyone holding the interpolator object. Worth flagging in the release notes as a breaking change, not only as a fix.
2. **The updated ELFCAR reference value** (`0.0918471` → `0.1167966`, ~27%) is self-consistent with the new code but not independently anchored. Cross-checking that one point against VESTA or a direct nearest-grid readout would turn the test into a real oracle rather than a snapshot of current behavior.

Nit: `xpoints`/`ypoints`/`zpoints` are no longer used by the interpolation path and are kept only for backwards compatibility — a one-line comment saying so would help the next reader.

shyuep:
Thanks!

PR #128, opened 2026-08-15
Add compatibility.md to document/announce breaking changes

### LINKED ISSUE #120: How to announce breaking changes?
Previously we had the [docs/compatibility.md](https://github.com/materialsproject/pymatgen/blob/main/docs/compatibility.md) to record breaking changes, and now docs are all moved the to `pymatgen` repo. Wondering what should be the current channel to communicate breaking changes?

DanielYang59:
close https://github.com/materialsproject/pymatgen-core/issues/120

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,22 @@
+## Summary
+
+<!-- Briefly describe what this PR does and why. Link any related issues/PRs. -->
+
+## Checklist
+
+- [ ] Tests for the affected code pass locally (e.g. `uv run pytest tests/core -k lattice`
+      for a lattice change — the full suite runs in CI)
+- [ ] Lint passes: `uv run ruff check . && uv run ruff format --check .`
+
+## Breaking changes?
+
+- [ ] This PR contains **breaking changes** (removals/renames, signature or
+      behavior changes, changed defaults/output, dropped Python versions, ...).
+      If so, prefix the title with `[breaking]` and describe the change and
+      migration steps under `## Breaking Changes` below.
+
+## Breaking Changes
+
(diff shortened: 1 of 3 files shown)


shyuep (commented):

Nice addition — a machine-maintained `COMPATIBILITY.md` driven off `[breaking]` PR titles is a good fit for the split repo, and the existing entries backfill cleanly. A few points:

1. `_extract_breaking_notes` terminates on `^##\s`, which does not match `###`/`####` subheadings, so any `###` section following `## Breaking Changes` gets swallowed into the notes. Terminating on `^#{1,6}\s` (excluding the start line) would be safer.
2. The PR template leaves an HTML comment inside the `## Breaking Changes` section. If a contributor fills in the section but leaves the `<!-- Optional. ... -->` block, it is copied verbatim into `COMPATIBILITY.md`. Worth stripping `<!-- ... -->` in `_extract_breaking_notes`.
3. `update_changelog(version: str | None = None)` — `_compatibility_section` interpolates `version` directly into `f"### v{version}"`, so a `None` version would emit `### vNone`. Please guard or assert non-None before writing.
4. Wording nit: the final `print` says "No `breaking`-labeled PRs", but detection is by title prefix, not by GitHub label. Same for the `_compatibility_section` docstring.
5. `ctx.run("open docs/CHANGES.md")` → `ctx.run("open CHANGES.md")` is a correct fix for this repo but unrelated to the stated purpose of the PR — fine to keep, just call it out in the description.

Only the `lint` job ran (pass); no test job, which is expected for a docs/`tasks.py`-only change. Recommend a dry-run of `invoke update-changelog --dry-run` against the last tag to confirm the insertion point and formatting before merge.

PR #129, opened 2026-08-17
Keep phonon thermodynamic integrals finite when the DOS samples omega = 0

Divide-By-0:
## Summary

`PhononDos.cv`, `entropy`, `internal_energy` and `helmholtz_free_energy` return `nan` — silently, no exception — whenever the frequency grid samples `omega = 0` exactly.

All four integrate a Bose-Einstein weight against the DOS from the first non-negative frequency (`_positive_frequencies`, which includes 0). Each weight is singular there — `x**2/sinh(x)**2` and `omega*coth(omega/2kT)` are `0/0`, `log(2*sinh(x))` is `log 0` — so the integrand picks up a `nan` at the first sample and `np.trapezoid` propagates it to the result.

Grids that start at exactly 0 are not exotic. phonopy writes one whenever a DOS range is requested:

```console
$ phonopy --dos --dos-range="0 16 0.1"      # first row of total_dos.dat is 0.0000000 0.0000000
```

and `np.linspace(0, f_max, n)` gives the same thing, so `pymatgen.io.phonopy.get_ph_dos` hands `PhononDos` such a grid routinely. Reproducer, using phonopy's own bundled Si example (`example/Si`, 20x20x20 mesh) round-tripped through pymatgen's own reader:

```python
from pymatgen.io.phonopy import get_ph_dos

dos = get_ph_dos("total_dos.dat")
dos.frequencies[0], dos.densities[0]   # 0.0, 0.0  <- density vanishes, so the limit is finite
dos.cv(300)                            # nan  (before) -> 39.8298  (after)
dos.entropy(300)                       # nan  (before) -> 40.6312  (after)
dos.internal_energy(300)               # nan  (before) -> 18423.1  (after)
dos.helmholtz_free_energy(300)         # nan  (before) ->  6233.8  (after)
```

phonopy's own `run_thermal_properties` on the same force constants gives `C_v = 39.8354` and `S = 40.6827 J/K/mol-cell`, so the recovered values agree to 0.01% and 0.13% — the residual is ordinary DOS-sampling error, which is the point: nothing about `omega = 0` was ever physically singular here, the density vanishes there.

### Fix

Evaluate each weight only where `omega > 0` and substitute the analytic `omega -> 0` limit at the zero sample, so the integration domain is left intact:

- `C_v`: `x**2/sinh(x)**2 -> 1`
- `U`: `omega*coth(omega/2kT) -> 2kT`
- `S`, `F`: the weights have no finite limit — they diverge logarithmically. The divergence is integrable, so dropping the `omega = 0` sample converges to the correct integral, and it contributes exactly nothing for the harmonic lattice DOS these functions model (`g ~ omega**2` in 3D). A DOS with `g(0) > 0` — a VDOS from the velocity autocorrelation of a diffusive MD trajectory, say — is outside that model, so it warns rather than quietly under-integrating (per review).

Grids that do not sample `omega = 0` take an identical code path to before — the existing NaCl reference values in `test_thermodynamic_functions` are unchanged, byte for byte.

`zero_point_energy` was already immune (no division) and is untouched.

### Test

`test_thermodynamic_functions_with_zero_frequency_sample` builds a Debye-like DOS on `np.linspace(0, 20, 401)`, asserts all four functions are finite at 10 K and 300 K, and asserts they match the same DOS with the `omega = 0` sample dropped to within 1e-3 relative — i.e. the zero sample now contributes only its trapezoid panel, not a `nan`. It fails on `main` with `assert np.isfinite(nan)` and passes with the fix.

`test_thermodynamic_functions_warn_on_finite_zero_frequency_density` covers the `g(0) > 0` case: `C_v` and `U` stay silent and exact, `S` and `F` warn, and refining the grid 201 -> 6401 points moves `S` by 1.2% and keeps shrinking — i.e. the dropped sample converges rather than diverging.

## Human intent

Rephrased from the user's request:

- While auditing reproducibility of materials-science codes (phonopy, hiPhive, BoltzTraP2, CGCNN), review pymatgen for defects that those workflows would actually hit, and upstream a fix.
- Prefer a real, reproducible correctness bug with a fail-to-pass regression test over a documentation or style change.

## Checklist

- [x] Tests for the affected code pass locally (`uv run pytest tests/phonon tests/io/test_phonopy.py` — 75 passed, 21 skipped for the missing optional phonopy dep)

## Breaking changes?

- [ ] This PR contains **breaking changes**.

No behavior change for any grid that does not sample `omega = 0`; grids that do previously produced `nan`.

### DIFF
--- a/src/pymatgen/phonon/dos.py
+++ b/src/pymatgen/phonon/dos.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import warnings
 from typing import TYPE_CHECKING, Literal, NamedTuple
 
 import numpy as np
@@ -28,7 +29,7 @@ THZ_TO_J = const.value("hertz-joule relationship") * const.tera
 
 
 def _thermo_integrand(
-    freqs: NDArray, dens: NDArray, weight: Callable[[NDArray], NDArray], zero_freq_weight: float
+    freqs: NDArray, dens: NDArray, weight: Callable[[NDArray], NDArray], zero_freq_weight: float | None
 ) -> NDArray:
     """Build the integrand weight(freqs) * dens for a thermodynamic DOS integral.
 
@@ -44,18 +45,38 @@ def _thermo_integrand(
         freqs: non-negative frequencies, ascending.
         dens: densities corresponding to freqs.
         weight: the Bose-Einstein weight, evaluated only where freqs > 0.
-        zero_freq_weight: value used where freqs == 0. For C_v and U this is the exact
-            limit of the weight. For S and F the weight itself diverges logarithmically,
-            but the divergence is integrable and any physical phonon DOS vanishes as
-            omega -> 0 (g(omega) ~ omega^2), so the omega = 0 sample contributes nothing
-            and 0 is used.
+        zero_freq_weight: value used where freqs == 0, or None when the weight has no
+            finite limit there. C_v and U have exact limits (1 and 2kT). The S and F
+            weights instead diverge logarithmically; the divergence is integrable, so
+            dropping the omega = 0 sample converges to the correct integral, and it is
+            exactly zero for the harmonic lattice DOS these functions model, which
+            vanishes as g(omega) ~ omega^2. A DOS with g(0) > 0 -- e.g. a VDOS from the
+            velocity autocorrelation of a diffusive MD trajectory -- is outside that
+            model and would need the first panel resolved (or a two-phase treatment of
+            the diffusive component), so warn instead of quietly under-integrating.
 
     Returns:
         NDArray: the integrand, finite everywhere freqs and dens are finite.
     """
-    weights = np.full_like(freqs, zero_freq_weight, dtype=float)
     positive = freqs > 0
+    weights = np.zeros_like(freqs, dtype=float)
     weights[positive] = weight(freqs[positive])
+
+    if zero_freq_weight is None:
+        zero_freq_dens = dens[~positive]
+        if zero_freq_dens.size and np.any(zero_freq_dens != 0):
+            warnings.warn(
+                f"DOS has non-zero density {np.max(zero_freq_dens)} at omega = 0, where the entropy and "
+                "free-energy weights diverge logarithmically. The omega = 0 sample is dropped, so the result "
+                "is under-integrated by O(delta * log delta) in the frequency spacing. Refine the grid near "
+                "zero, or note that a DOS with g(0) > 0 (a diffusive VDOS, say) is outside the harmonic "
+                "oscillator model these functions assume.",
+                UserWarning,
+                stacklevel=3,
+            )
+    else:
+        weights[~positive] = zero_freq_weight
+
     return weights * dens
 
 
@@ -232,7 +253,7 @@ class PhononDos(MSONable):
             return wd2kt**2 / np.sinh(wd2kt) ** 2
 
         # x**2 / sinh(x)**2 -> 1 as x -> 0
-        cv = np.trapezoid(_thermo_integrand(freqs, dens, cv_weight, 1.0), x=freqs)
+        cv = float(np.trapezoid(_thermo_integrand(freqs, dens, cv_weight, 1.0), x=freqs))
         cv *= const.Boltzmann * const.Avogadro
 
         if structure:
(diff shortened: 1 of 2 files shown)


shyuep (commented):

The diagnosis is correct and the fix is the right shape: evaluate the Bose-Einstein weight only on `freqs > 0` and substitute the analytic limit at `omega = 0`, keeping the trapezoid domain intact. `x**2/sinh(x)**2 -> 1` and `omega*coth(omega/2kT) -> 2kT` are both right, and the refactor into a single `_thermo_integrand` helper removes four copies of the same bug. Tests cover both the finiteness requirement and agreement with the trimmed grid.

Three points:

1. **The `zero_freq_weight = 0.0` choice for `entropy`/`helmholtz_free_energy` is only valid when `g(0) = 0`.** The docstring justifies it with "any physical phonon DOS vanishes as `omega -> 0` (`g ~ omega^2`)", which holds for a harmonic lattice DOS but *not* for a VDOS obtained from the velocity autocorrelation function of an MD trajectory, where the diffusive contribution gives finite `g(0)`. Those are increasingly common inputs (MLIP-driven MD). In that case the integral genuinely diverges logarithmically and the patch silently returns a finite, under-estimated `S`/`F` instead of `nan` — arguably a worse failure mode than the current one. Suggest warning when `freqs[0] == 0 and dens[0] != 0`, e.g. in `_thermo_integrand` when `zero_freq_weight` is the sentinel-zero case.

2. **Minor inconsistency:** `internal_energy` now wraps the result in `float(...)` while `cv`, `entropy` and `helmholtz_free_energy` do not. Either apply uniformly or drop it.

3. **`np.full_like(freqs, ...)`** is fine, but note `weight` is still called on a copy-slice `freqs[positive]`; for large grids this is two extra allocations per call. Not a concern at typical DOS sizes — just flagging in case `cv` is called in a temperature loop.

**CI:** no checks have run on this PR at all (`statusCheckRollup` is empty), which for a first-time contributor usually means workflows are awaiting maintainer approval. Recommend approving the run before merge — the numerical assertions in `test_thermodynamic_functions_with_zero_frequency_sample` are the whole point of the patch and are currently unverified.

Otherwise this looks ready pending point 1.

Divide-By-0:
Thanks — all three addressed in eb6a3c9.

**1. `g(0) > 0` (the diffusive VDOS case).** Agreed on the failure mode, with one correction to the premise: the integral does not diverge. `x*coth(x) - log(2*sinh(x)) ~ 1 - log(2x)` is integrable against a finite `g(0)`, so dropping the `omega = 0` sample converges to the correct value rather than to a wrong one. Refining a flat `g = 1` DOS on `[0, 10]` bears that out:

| grid | delta | S(300) | F(300) |
|---|---|---|---|
| 201 | 0.050 | 128.477 | -11932.5 |
| 401 | 0.025 | 129.202 | -12118.8 |
| 1601 | 0.00625 | 129.818 | -12280.1 |
| 6401 | 0.00156 | 129.999 | -12328.5 |
| 25601 | 0.00039 | 130.051 | -12342.7 |

Converging, with the increment shrinking about 4x per 4x refinement — the `O(delta * log delta)` you would expect from an integrable log endpoint singularity.

That still leaves your real point standing: on a coarse grid the underestimate is ~1.2% in S and ~3.3% in F, and a DOS with `g(0) > 0` is outside the harmonic-oscillator model these functions assume in the first place (the diffusive component wants a 2PT-style treatment, not a quantum oscillator). So `zero_freq_weight` is now `float | None`, where `None` means "no finite limit here" and triggers a warning when the density at `omega = 0` is non-zero. That also kills the sentinel ambiguity you flagged — `0.0` and "the weight diverges" are no longer spelled the same way. `C_v` and `U` pass their exact limits (`1` and `2kT`) and never warn.

Covered by `test_thermodynamic_functions_warn_on_finite_zero_frequency_density`, which also asserts the convergence above, plus a `simplefilter("error")` guard on the harmonic case so a vanishing `g(0)` stays silent.

**2. `float(...)` inconsistency.** Applied uniformly to all four. It was load-bearing in `internal_energy` only because the extra `/ 2` made mypy infer `ndarray | Any` against the declared `-> float`; making it uniform is the better fix.

**3. Allocations.** Left as is. The masked evaluation is what guarantees no `nan`/`inf` is ever materialized — computing the weight over the full grid and repairing afterwards would reintroduce exactly the failure this patch removes — and the two temporaries are `O(n_grid)` against a `trapezoid` of the same order, so a temperature loop pays a constant factor on an already-linear pass.

**CI:** still no checks — as you note, this needs a maintainer to approve the workflow run for a first-time contributor. Happy to have it exercised; the assertions are the substance of the patch. Locally: `uv run pytest tests/phonon tests/io/test_phonopy.py` gives 75 passed / 21 skipped (skips are the missing optional phonopy dep), `ruff check` and `ruff format --check` are clean repo-wide, and mypy is unchanged from baseline.

shyuep:
Thanks!

PR #131, opened 2026-08-20
Fix CHGCAR augmentation corruption from doubled newlines

hanaol:
## Summary

`VolumetricData.write_file` corrupts CHGCARs whose `data_aug` is in the legacy
list-of-lines format.

The legacy branch in `write_aug` appends a newline to lines that already end with
one:

```python
if isinstance(aug_data, (list, tuple)):
    for line in aug_data:
        file.write(f"{line}\n")
```

Legacy `data_aug` entries come straight from `for line in file` (the pre-rewrite
`parse_file` appended `original_line`), so they already carry their `\n`.

Thus, VASP prints

```
RD_RHO_PAW: ion            2 data corrupt
WARNING: PAW occupancies are missing on CHGCAR
```

to stdout when running with `ICHARG=1` and initializing with a CHGCAR file.

### Origin

Before the augmentation-parsing rewrite, this branch read:

```python
file.write("".join(data))
```

which was symmetric with a parser that stored lines *with* their trailing
newline. The rewrite replaced it with a per-line write that assumes *stripped*
lines. The new dict path is self-consistent, but the legacy branch is now wrong.

Bisected against released artifacts: correct in `pymatgen-core` 2026.4.16 and in
every earlier release checked back to `pymatgen` 2024.8.9; broken from 2026.5.17
onward.

### Fix

Write the newline only when the line does not already have one. Output is
byte-identical for already-stripped lines and for non-`str` entries, so callers
that worked around this by pre-stripping their `data_aug` are unaffected.

### A note on existing coverage

`test_write_preserves_legacy_aug_lines` already covers this branch, but its
fixture uses lines *without* trailing newlines — the one convention the broken
code handles correctly — so it passed before and after this change. That is why
the defect shipped despite nominal coverage. The added test uses the real legacy
format and asserts that no blank line appears inside an augmentation block; it
fails on `main` and passes with this change.

## Checklist

- [x] Tests for the affected code pass locally:
      `pytest tests/io/vasp/test_outputs.py tests/io/test_common.py`
      → 195 passed, 0 failed
- [x] Lint passes: `ruff check .` and `ruff format --check .`

## Breaking changes?

- [ ] This PR contains **breaking changes**

### DIFF
--- a/src/pymatgen/io/vasp/outputs.py
+++ b/src/pymatgen/io/vasp/outputs.py
@@ -4026,8 +4026,10 @@ class VolumetricData(BaseVolumetricData):
             aug_data: Any = self.data_aug.get(data_type, {})
 
             if isinstance(aug_data, (list, tuple)):
+                # Legacy lines already carry their trailing newline; a second one corrupts the block.
                 for line in aug_data:
-                    file.write(f"{line}\n")
+                    text = str(line)
+                    file.write(text if text.endswith("\n") else f"{text}\n")
                 return
 
             if not isinstance(aug_data, dict):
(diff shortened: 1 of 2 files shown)


shyuep:
Thanks!

PR #132, opened 2026-08-22
Fix `SpacegroupAnalyzer.get_point_group_symbol()` for cells with non-conventional basis orientations

kavanase:
## Summary
`SpacegroupAnalyzer.get_point_group_symbol()` currently determines the point group from `spglib`'s `dataset.rotations` (from `spglib.get_pointgroup()`). The `spglib` dataset only returns operations expressible as integer matrices in the input cell basis, so for structures described in a (super)cell whose basis vectors are not aligned with the conventional cell axes, part of the crystal's point operations are missing from `rotations`, and the method returns a lower symmetry point group (the expressible subgroup) as a result, which is inconsistent with `get_space_group_symbol()`, which instead uses spglib's basis-independent type identification.

MWE (pristine diamond in a rotated-cube 3x supercell):
```python
from pymatgen.core import Lattice, Structure
from pymatgen.symmetry.analyzer import SpacegroupAnalyzer

diamond = Structure.from_spacegroup("Fd-3m", Lattice.cubic(3.5597), ["C"], [[0, 0, 0]])
sc = diamond.copy()
sc.make_supercell([[2, 2, -1], [-1, 2, 2], [2, -1, 2]])  # 3a0 cube, edges along <221> axes
sga = SpacegroupAnalyzer(sc, symprec=0.01)
sga.get_space_group_symbol()  # "Fd-3m" (correct)
sga.get_point_group_symbol()  # "-3m"   (wrong; should be "m-3m")
```

Such mis-aligned supercells arise routinely from cell-shape optimisation in defect/phonon/MD workflows.

The fix is to directly return `dataset.pointgroup`, which `spglib` computes from the space-group type (basis-independent, as with `get_space_group_symbol()` -- now self-consistent) and which is already held in self._space_group_data. 

The `SpglibDataset.pointgroup` field is available throughout pymatgen's supported `spglib` range (>=2.5). A regression test covering both aligned and rotated diamond supercells (this MWE) is included.

## Checklist

### DIFF
--- a/src/pymatgen/symmetry/analyzer.py
+++ b/src/pymatgen/symmetry/analyzer.py
@@ -197,11 +197,8 @@ class SpacegroupAnalyzer:
         Returns:
             Pointgroup: Point group for structure.
         """
-        rotations = self._space_group_data.rotations
-        # passing a 0-length rotations list to spglib can segfault
-        if len(rotations) == 0:
-            return "1"
-        return spglib.get_pointgroup(rotations)[0].strip()  # type:ignore[index]
+        # determined by spglib from the space group type, and so is basis-independent:
+        return self._space_group_data.pointgroup
 
     def get_crystal_system(self) -> CrystalSystem:
         """Get the crystal system for the structure, e.g. (triclinic, orthorhombic,
--- a/tests/symmetry/test_analyzer.py
+++ b/tests/symmetry/test_analyzer.py
@@ -91,6 +91,19 @@ class TestSpacegroupAnalyzer(MatSciTest):
         assert self.sg.get_point_group_symbol() == "mmm"
         assert self.disordered_sg.get_point_group_symbol() == "4/mmm"
 
+    def test_get_pointgroup_unaligned_supercell(self):
+        # point group determination is basis-independent, consistent with the space group type, including
+        # for supercells whose basis vectors are not aligned with the conventional cell axes
+        # (where some point operations are not expressible as integer matrices in the cell basis) --
+        # previously a failure case, when point-group determination was based on spglib rotations dataset
+        diamond = Structure.from_spacegroup("Fd-3m", Lattice.cubic(3.5597), ["C"], [[0, 0, 0]])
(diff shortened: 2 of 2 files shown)


shyuep (commented):
**[Automated PR review generated by Claude]**

Clean, well-motivated fix. Using `dataset.pointgroup` (derived from the space-group type) instead of `spglib.get_pointgroup(dataset.rotations)` correctly makes `get_point_group_symbol()` basis-independent and self-consistent with `get_space_group_symbol()`. The MWE (rotated 3a0 diamond supercell returning -3m instead of m-3m) clearly demonstrates the bug.

Points checked:
- The removed 0-length-rotations segfault guard is no longer needed since `spglib.get_pointgroup()` is no longer called; failure modes for a null spglib dataset are unchanged (`_space_group_data` access was already the first statement).
- New `test_get_pointgroup_unaligned_supercell` covers both aligned and rotated supercells.
- All CI green (lint + 4 test matrices incl. lowest-direct and py3.14).

Note this is technically behavior-changing for any structure where the two methods previously disagreed — worth a mention in the release notes. LGTM.

shyuep (commented):
**🤖 Automated PR review generated by Claude** (posted via @shyuep's scheduled review routine)

Good fix. Using `dataset.pointgroup` (derived from the space-group type) instead of `spglib.get_pointgroup(dataset.rotations)` makes `get_point_group_symbol()` basis-independent and self-consistent with `get_space_group_symbol()` — the rotated-supercell MWE is convincing, and the regression test covers both aligned and rotated cases. Also nicely removes the 0-length-rotations segfault guard, since `pointgroup` is always populated when the dataset exists.

Two minor points, neither blocking: (1) this silently changes returned symbols for previously mis-detected structures — worth a mention in the release notes as a behavior fix; (2) if `_space_group_data` can be `None` for pathological inputs (symmetry detection failure), the old path and new path both fail upstream, so no new risk.

All CI green. LGTM.

shyuep:
Thanks. I agree with the fix.

PR #133, opened 2026-08-24
Add `symmetric` option to `group_structures`, pass through to `sm.fit()` and test

kavanase:
## Summary

This is a small PR to add the `symmetric` option to `StructureMatcher.group_structures()`, which is then passed through to the underlying `StructureMatcher.fit()` usage.

Summary with the help of Claude:
`StructureMatcher.fit` is directional; `symmetric=False` by default, and its docstring says so. But `group_structures` consumes it as if it were an equality test: it pops a reference and always passes that reference as `struct1`. So which structure happens to come first decides the grouping.

This repo already keeps a structure pair for the asymmetry, used by `test_fit`:
```python
# src/pymatgen/util/testing.py:36
VASP_IN_DIR: str = f"{TEST_FILES_DIR}/io/vasp/inputs"
```

```python
sm = StructureMatcher(comparator=ElementComparator(), ltol=0.6, stol=0.6, angle_tol=6)
s1 = Structure.from_file(f"{VASP_IN_DIR}/POSCAR_fit_symm_s1")
s2 = Structure.from_file(f"{VASP_IN_DIR}/POSCAR_fit_symm_s2")

len(sm.group_structures([s1, s2]))                   # 1 group
len(sm.group_structures([s2, s1]))                   # 2 groups   <-- same inputs
len(sm.group_structures([s1, s2], symmetric=True))   # 2 groups
len(sm.group_structures([s2, s1], symmetric=True))   # 2 groups
```

Using `test-files/io/vasp/inputs/POSCAR_fit_symm_s1` and `_s2`, from `tests/core/test_structure_matcher.py:529-537`.

This PR threads a `symmetric` kwarg through to the `fit` call so callers that need an order-independent
grouping can ask for one. It defaults to `False`, so no existing caller changes behaviour, and it is
ignored under `anonymous=True` since `fit_anonymous` has no symmetric mode. One test added.

**Motivation:** `MaterialsProjectDFTMixingScheme` groups GGA and r2SCAN structures this way, and the
asymmetry makes its phase diagrams depend on `PYTHONHASHSEED` — see materialsproject/pymatgen#3113 and
the follow-up PR (I'll be raising now) to materialsproject/pymatgen#4703, which passes `symmetric=True` here. Over the 120 permutations of the pre-group in that bug, the current default yields 3 distinct groupings and `symmetric=True` yields 1, at 1.04x runtime for the full pipeline.

Note this does not make greedy grouping order-independent in general — a symmetric relation can still be non-transitive — but it removes the directional class of it at least.

## Checklist

## Breaking changes?

- [ ] This PR contains **breaking changes** (removals/renames, signature or

### DIFF
--- a/src/pymatgen/core/structure_matcher.py
+++ b/src/pymatgen/core/structure_matcher.py
@@ -840,7 +840,7 @@ class StructureMatcher(MSONable):
 
         return None
 
-    def group_structures(self, s_list, anonymous=False):
+    def group_structures(self, s_list, anonymous=False, symmetric=False):
         """
         Given a list of structures, use fit to group
         them by structural equality.
@@ -848,6 +848,13 @@ class StructureMatcher(MSONable):
         Args:
             s_list ([Structure]): List of structures to be grouped
             anonymous (bool): Whether to use anonymous mode.
+            symmetric (bool):
+                Whether a pair must match in both directions to be
+                grouped. Defaults to `False`. `fit()` is not symmetric
+                and the reference below is always passed as `struct1`,
+                so by default the grouping depends on the order of
+                `s_list`. Ignored if `anonymous` is `True`, since
+                `fit_anonymous` has no symmetric mode.
 
         Returns:
             A list of lists of matched structures
@@ -890,7 +897,7 @@ class StructureMatcher(MSONable):
                     )
                 else:
                     inds = filter(
-                        lambda i: self.fit(refs, unmatched[i][1], skip_structure_reduction=True),
+                        lambda i: self.fit(refs, unmatched[i][1], symmetric=symmetric, skip_structure_reduction=True),
                         list(range(len(unmatched))),
                     )
                 inds = list(inds)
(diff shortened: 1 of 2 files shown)


shyuep (commented):
**🤖 Automated PR review generated by Claude** (posted via @shyuep's scheduled review routine)

Clean, minimal change. The order-dependence of `group_structures` from directional `fit()` is a real and long-standing footgun (cf. materialsproject/pymatgen#3113), and threading `symmetric` through with a `False` default preserves existing behavior. Docstring correctly notes the kwarg is ignored under `anonymous=True` and that greedy grouping remains non-transitive in general.

Test reuses the existing asymmetric POSCAR pair and covers both orderings — good. All CI green.

One minor note: `symmetric=True` roughly doubles the `fit` cost per pair; the 1.04x pipeline figure in the description is reassuring, so no action needed. LGTM.

shyuep:
Thanks.

PR #134, opened 2026-08-28
[breaking] Do not record the identity as a rotation axis in PointGroupAnalyzer

HiroYokoyama:
## Summary

I found an issue in the symmetry analyzer: when a molecule whose only rotational symmetry is a single C2 axis (e.g., tetramethylhydrazine) is analyzed, sch_symbol reports it as D2, while get_symmetry_operations() returns only 2 operations. D2 is a group of order 4, so this is a self-contradiction.
I then used Claude Opus 5 to analyze the issue, and found that _check_rot_sym() registers a 360° rotation (the identity, which is valid about any axis) in rot_sym as a rotation axis, so the condition len(rot_sym) >= 2 in _proc_sym_top() routes the molecule to the dihedral branch.

Measured data:
```
C2 rotation residual about each principal axis (max atom displacement):
   axis [ 0.9866  0.122 -0.1085]   residual = 1.4231 A
   axis [-0.1375  0.2628 -0.955 ]  residual = 1.4232 A
   axis [-0.088   0.9571  0.2761]  residual = 0.0001 A
Best over 40000 random axes at least 10 deg away from that axis: 0.7332 A

                      before      after
tetramethylhydrazine  D2 / 2 ops  C2 / 2 ops
water C2v/4, methane Td/24, ammonia C3v/6, allene D2d/8,
benzene D6h/24, H2O2 C2/2, ethene D2h/8, HCl C*v/1, CO2 D*h/2   (unchanged)
```

This PR stops the loop in _check_rot_sym() at 2, and also removes the if len(self.rot_sym) > 0: guard in _proc_sym_top() so that the perpendicular C2 search always runs; that guard was only ever satisfied by the phantom entry, so fixing the loop alone would make these molecules report as C1 instead.

This PR also adds a regression test for tetramethylhydrazine in tests/symmetry/test_analyzer.py; the test fails on main with assert 'D2' == 'C2'.

## Checklist

- [x] Tests for the affected code pass locally: pytest tests/symmetry (92 passed / 5 skipped, before the patch 91/5)
- [x] Lint passes: `ruff check` / `ruff format --check` passed

## Breaking changes?

- [x] This PR contains **breaking changes** (removals/renames, signature or

## Breaking Changes

This may be a breaking change:
sch_symbol changes for molecules whose unique axis of inertia carries no rotation of its own but which have one perpendicular C2 axis: they were reported as D2 and are now reported as C2. get_symmetry_operations() and get_pointgroup() are unaffected — they already returned the 2-operation group. Code that compared sch_symbol against the previous (incorrect) value must be updated; nothing else is affected.

### DIFF
--- a/src/pymatgen/symmetry/analyzer.py
+++ b/src/pymatgen/symmetry/analyzer.py
@@ -1183,8 +1183,8 @@ class PointGroupAnalyzer:
         unique_axis = self.principal_axes[ind]
         self._check_rot_sym(unique_axis)
         logger.debug(f"Rotation symmetries = {self.rot_sym}")
-        if len(self.rot_sym) > 0:
-            self._check_perpendicular_r2_axis(unique_axis)
+        # A perpendicular C2 can exist even when the unique axis carries none; that molecule is C2, not C1.
+        self._check_perpendicular_r2_axis(unique_axis)
 
         if len(self.rot_sym) >= 2:
             self._proc_dihedral()
@@ -1306,7 +1306,8 @@ class PointGroupAnalyzer:
         """
         min_set = self._get_smallest_set_not_on_axis(axis)
         max_sym = len(min_set)
-        for idx in range(max_sym, 0, -1):
+        # Stop at 2: a 360-degree rotation is the identity and is valid about any axis.
+        for idx in range(max_sym, 1, -1):
             if max_sym % idx != 0:
                 continue
             op = SymmOp.from_axis_angle_and_translation(axis, 360 / idx)
--- a/src/pymatgen/symmetry/analyzer.py
+++ b/src/pymatgen/symmetry/analyzer.py
@@ -1341,6 +1341,8 @@ class PointGroupAnalyzer:
         if len(self.rot_sym) == 0:
             logger.debug("Accidental spherical top!")
             self._proc_sym_top()
+            # It assigns sch_symbol itself, and rot_sym may still be empty afterwards.
+            return
         main_axis, rot = max(self.rot_sym, key=lambda v: v[1])
         if rot < 3:
             logger.debug("Accidental spherical top!")
--- a/tests/symmetry/test_analyzer.py
+++ b/tests/symmetry/test_analyzer.py
@@ -851,6 +851,37 @@ class TestPointGroupAnalyzer(MatSciTest):
         pg_analyzer = PointGroupAnalyzer(mol)
         assert pg_analyzer.sch_symbol == "Ih"
 
+    def test_symmetric_top_without_rotation_about_the_unique_axis(self):
+        """A single C2 perpendicular to the unique axis is C2, not D2."""
+        mol = Molecule(
+            ["N", "C", "C", "N", "C", "C"] + ["H"] * 12,
+            [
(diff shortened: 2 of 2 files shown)


shyuep (commented):
**Automated PR review generated by Claude** (posted by @shyuep's review bot).

Diagnosis and fix are correct. `idx=1` in `_check_rot_sym` registered a 360° rotation (the identity) as a rotation axis about any axis, so `len(rot_sym) >= 2` misrouted single-perpendicular-C2 molecules to `_proc_dihedral` (D2 symbol with only 2 ops — self-contradictory). Stopping the loop at 2 and unconditionally running `_check_perpendicular_r2_axis` is the right pair of changes; either alone would be wrong (C1 regression), as the PR notes.

Two points to verify before merge:

1. `rot_sym` can now be empty after `_check_rot_sym` where the phantom identity entry previously guaranteed at least one element. Please audit the other callers (`_proc_asym_top`, `_proc_cyclic`, spherical-top path) for any `self.rot_sym[0]` access or `len(rot_sym) > 0` logic that implicitly relied on the identity entry. The passing test suite suggests this is fine, but an explicit check is warranted since the change is in a shared helper.
2. The body states this may change `sch_symbol` for affected molecules but the breaking-change box is unticked — per repo convention, tick it and prefix the title with `[breaking]`.

Test is well-constructed (regression on Schoenflies symbol, operation count, and `rot_sym` order). CI is green. LGTM modulo point 1.

HiroYokoyama:
Point 1 was real. `_proc_sph_top`'s `len(self.rot_sym) == 0` fall-back to
`_proc_sym_top()` had no `return`, so it fell through into
`max(self.rot_sym, ...)` — safe only while the phantom identity guaranteed a
non-empty `rot_sym`. On an accidental spherical top with no rotational
symmetry: `D1` before, `ValueError` with this PR alone, `C1` with the `return`.

Pushed 191f39b: the `return` plus a regression test. The other consumers
(`_proc_asym_top`, `_proc_cyclic`, `_proc_dihedral`, `_find_mirror`) are
unaffected.

Point 2: breaking-change box ticked, title prefixed.

93 passed, 5 skipped; ruff clean.

shyuep (commented):
**Automated PR review generated by Claude** (posted by @shyuep's review bot).

Re-review of 191f39b7 (new commit since last review):

Both points from the previous review are addressed. The early `return` in `_proc_sph_top` after the accidental-spherical-top fallback fixes the `ValueError` that `max(self.rot_sym, ...)` would raise on an empty `rot_sym` — the last unguarded access. `_proc_dihedral`/`_proc_cyclic` are only reached via the `len(rot_sym) >= 2` / `== 1` branches and `_proc_asym_top` routes `len == 0` to `_proc_no_rot_sym`, so the audit is complete. The new `test_accidental_spherical_top_without_rotation_symmetry` regression test covers exactly this path, and the title now carries the `[breaking]` prefix.

One note: CI has not run on the latest commit (workflows pending approval). Approve the run and confirm green before merge. Otherwise LGTM.

shyuep:
Thanks for the fix. Admittedly the logic of point group analyzer has not been checked robustly in the past.

PR #136, opened 2026-08-29
Avoid ``UFloat`` warning spam with mixed MP PDs

kavanase:
## Summary

Follow-up to #4386 / #4400 / #4406, which left one case open: an energy adjustment that is present with `uncertainty == 0.0`.

`uncertainties>=3.2` warns inside the `ufloat()` constructor whenever `std_dev == 0` (`uncertainties/core.py:1024`). Materials Project entries carry `ConstantEnergyAdjustment("MP GGA(+U)/r2SCAN mixing adjustment", uncertainty=0.0)`
on every r2SCAN entry, so `ComputedEntry.correction` re-fires the warning on every `.energy` access.

MWE:
```python
from mp_api.client import MPRester
from pymatgen.analysis.phase_diagram import PhaseDiagram

with MPRester() as mpr:
    entries = mpr.get_entries_in_chemsys(["Cs", "Ti", "I"])
PhaseDiagram(entries)   # UserWarning: Using UFloat objects with std_dev==0 ...
```

`ufloat` is not needed in either property (correction or uncertainty) — correction only returns `.nominal_value`, and the uncertainty is a plain quadrature sum:

- `correction` -> `math.fsum(ea.value for ea in self.energy_adjustments)`
- `correction_uncertainty` -> `math.hypot(*(...)) or np.nan`  (i.e. `sum(uncertainty**2)**0.5`; quadrature sum)

Also folds in the same warning family in `analysis/reaction_calculator.py`:
`min(float, UFloat)` fell through to the deprecated `AffineScalarFunc.__lt__`, emitting `FutureWarning` from `test_calculated_reaction_energy_uncertainty`. Now compares on `nominal_value`, which also removes the `float("inf")` sentinel.

All existing tests pass, plus new `test_zero_uncertainty_adjustment_doesnt_warn` here. `ComputedEntry.energy` is also now ~20x faster (20k accesses: 0.059 s -> 0.003 s)(e.g. `PhaseDiagram` construction, which uses many `.energy` accesses).

## Checklist

- [ ] Lint passes: `uv run ruff check . && uv run ruff format --check .`

## Breaking changes?

- [ ] This PR contains **breaking changes** (removals/renames, signature or

### DIFF
--- a/src/pymatgen/analysis/reaction_calculator.py
+++ b/src/pymatgen/analysis/reaction_calculator.py
@@ -10,7 +10,7 @@ from typing import TYPE_CHECKING, overload
 import numpy as np
 from monty.fractions import gcd_float
 from monty.json import MontyDecoder, MSONable
-from uncertainties import UFloat, ufloat
+from uncertainties import UFloat, nominal_value, ufloat
 
 from pymatgen.core.composition import Composition
 from pymatgen.core.entries import ComputedEntry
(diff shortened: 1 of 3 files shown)


shyuep:
Thanks!

PR #137, opened 2026-09-11
Restore fail-fast error in AseAtomsAdaptor when ASE is missing

shyuep:
## Summary

Without ASE installed, `Structure.to_ase_atoms()` failed deep inside `AseAtomsAdaptor.get_atoms` with an obscure error instead of telling the user to install ASE:

```
AttributeError: 'MSONAtoms' object has no attribute 'set_array'
```

Root cause: 35d31b180, 4b77a8ec7 and d5d50fb03c (Feb 2026) replaced the stub `Atoms.__init__` `raise NO_ASE_ERR` with `pass`, dropped the `NO_ASE_ERR` guard at the top of `get_atoms`, and deleted `test_no_ase_err`. `MSONAtoms` then silently inherited from the do-nothing stub, and `NO_ASE_ERR` was defined but never raised anywhere.

## Changes

- `src/pymatgen/io/ase.py`: stub `Atoms.__init__` raises `NO_ASE_ERR` again; `get_atoms` re-checks the sentinel before doing any work. Callers now get `PackageNotFoundError("AseAtomsAdaptor requires the ASE package. Use `pip install ase`")`.
- `tests/io/test_ase.py`: new `test_no_ase_err` that monkeypatches the module-level sentinel. The deleted test reloaded `pymatgen.io.ase` with `ase.atoms` blocked, which leaked a stubbed module into other tests in the same xdist worker and is the likely reason the guards were removed in the first place.

## Test plan

- [x] `ruff check` / `ruff format` clean; pre-commit (mypy, codespell) passes
- [x] `tests/io/test_ase.py` + `tests/core/test_structure.py` with ASE installed: 194 passed, 11 skipped
- [x] Same files with `ase` blocked in-process (`-n0`), mimicking the CI legs without the `optional` extra: 154 passed, 23 skipped; module import and `MSONAtoms` class definition still work
- [x] Original repro with ASE blocked now raises `PackageNotFoundError` from `to_ase_atoms()`, `MSONAtoms()` and `Atoms()`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


shyuep:

Correct diagnosis and a minimal fix. The regression is real: `NO_ASE_ERR` was defined but never raised after the Feb 2026 refactor, so the stub `Atoms` silently produced a crippled `MSONAtoms`. CI is green on all legs.

Three small things:

1. **`raise NO_ASE_ERR` re-raises a single module-level exception instance.** Every call raises the *same* `PackageNotFoundError` object, so `__traceback__` accumulates frames across calls and any `__context__`/`__cause__` set by an earlier raise persists. Harmless in practice but it makes repeated failures noisy. `raise PackageNotFoundError(...)` from a module-level message constant, or `raise NO_ASE_ERR from None`, avoids it. (This matches the pre-regression code, so not a new problem — just worth fixing while you are here.)

2. **The guard covers `get_atoms` but not the other entry points.** `MSONAtoms.from_dict` calls `decode(...)`, which is `None` in the no-ASE branch, so it still fails with `TypeError: 'NoneType' object is not callable` rather than the informative error. `AseAtomsAdaptor.get_structure` takes an `Atoms` the caller could not have constructed, so it is fine. Adding the sentinel check to `from_dict` would close the loop for one more line.

3. **The test never exercises the stub.** Monkeypatching the module-level sentinel is the right call given the reload-leak you documented, but it means the `Atoms.__init__`/`MSONAtoms.__init__` path — the actual root cause — has no coverage; only the new `get_atoms` guard does. A second assertion on a locally-defined stub-equivalent, or a subprocess-isolated test (`pytest.mark.subprocess` style / `subprocess.run([sys.executable, "-c", ...])`), would cover it without polluting the xdist worker.

None of these block merge.

PR #138, opened 2026-09-11
tests: probe seekpath usability instead of gating on platform/Python version

Yi-111-a:
## Summary

`tests/symmetry/test_kpaths.py` and `tests/symmetry/test_kpath_hin.py` skip every seekpath-backed k-path test on Windows or Python ≥ 3.13 via a hard-coded gate:

```python
if sys.platform.startswith("win") or not (sys.version_info <= (3, 13)):
    pytest.skip(...)
```

Because of this gate plus the CI matrix, **these tests currently never execute on any CI job**:

| CI job | seekpath installed? | Why it skips |
|---|---|---|
| `windows-latest 3.13` (no extras) | no | `importorskip` |
| `macos-latest 3.13` `extras: optional` | yes | version gate |
| `ubuntu-latest 3.11` (no extras) | no | `importorskip` |
| `ubuntu-latest 3.14` `extras: optional` | yes | version gate |

The gate is stale. It was added on 2026-05-08 in `3643cdb4d` to work around a real
failure, but the matrix changed 11 days later in a way that made the Windows branch dead
code, while the version branch still suppresses coverage on the two jobs that *do*
install seekpath.

This PR replaces the platform/version gate with a runtime capability probe: a single
cubic P1 cell (three general-position sites) exercises `KPathSeek(...).get_kpoints()`,
so the tests run wherever seekpath actually works and skip (with an accurate reason)
wherever it does not. The probe only establishes that `KPathSeek` is callable — not
that every Bravais path works; that coverage stays in the seekpath test modules.

## Root cause of the original failure

The original breakage came from the upstream library, not from pymatgen:
`seekpath.hpkot.tools.get_dot_access_dataset()` did not handle a `None` dataset —
the failure signal emitted by `spglib < 2.7` — so a symmetry-detection failure surfaced as
`TypeError: types.SimpleNamespace() argument after ** must be a mapping, not NoneType`.

Fixed upstream in `materialscloud-org/seekpath#118` (merged 2026-01-29, released in 2.2.1).

A second, independent failure (module-reload pollution from
`test_err_msg_on_seekpath_not_installed`) was also suppressing the seekpath tests under
pytest-xdist; this PR restores the real modules in a `finally` block after the
simulation.

### Timeline (from `git log` and GitHub check runs)

| Date | Event | CI |
|---|---|---|
| 2026-05-08 | `e5439341f` replaces import-time check with a runtime probe | windows+macos **failure** |
| 2026-05-08 | `3643cdb4d` adds the platform/version gate | green |
| 2026-05-19 | `9f9254430` *"remove optional testing on Windows since it is super slow"* | — |
| 2026-08-30 | current `main` | seekpath tests skipped everywhere |

`uv.lock` has pinned `seekpath==2.2.1` / `spglib==2.7.0` since before the gate was added
and still does today, so the pinned environment the gate was written against is the one
running now.

## Changes

- `pymatgen.util.testing.seekpath_unusable_reason()`: single-cell probe of `KPathSeek`;
  unavailability detected via `kpath.get_path is None` (not exception-message text);
  in-cell fractional coords with `to_unit_cell=True`.
- `tests/symmetry/test_kpaths.py`, `tests/symmetry/test_kpath_hin.py`: drop the
  `sys.platform` / `sys.version_info` branch; skip via the shared probe instead.
- `tests/io/pwmat/test_inputs.py`: restore `kpath` / `pwmat.inputs` after the simulated
  absence so later seekpath tests are not poisoned under xdist.
- Skip reason text updated from "seekpath not usable on this platform/Python" to
  "seekpath unusable in this environment", which describes what is actually checked.
- `import sys` removed from the symmetry test files (no longer referenced).

## Test plan

Measured locally with `seekpath==2.2.1` / `spglib==2.7.0`:

- [x] **Windows, Python 3.13**, extras `optional`: seekpath tests pass (no longer skipped by platform/version gate)
- [x] **Windows, Python 3.14**, extras `optional`: same
- [x] Full sweep of all 230 space groups (`TestKPathSeek::test_kpath_generation`) passes on
      Windows 3.13 **and** 3.14
- [x] **Windows, Python 3.13, no extras** (mirrors the CI Windows leg): skip reasons correct — no regression when seekpath is absent
- [x] Poisoned trio `test_inputs` + `test_kpath_hin` + `test_kpaths` in one process: pollution fixed by `finally` restore
- [x] `pytest tests/symmetry`: `100 passed`
- [x] `ruff check` + `ruff format --check`: clean
- [x] Fork CI matrix green across all five legs (ubuntu/macos/windows)

### Known limitation / reviewer request

I could only run **Windows x86-64** locally for early verification; Linux and macOS were
confirmed on the fork CI matrix after the pollution fix. The end-to-end probe catches
environments where `KPathSeek` is broken: if the probe fails for a known-unavailability
signature the tests skip; anything else re-raises.


Yi-111-a:
The verification run on my fork changed the diagnosis in an important way, so recording it here.

**The probe alone is not sufficient — my fork's CI proved it.** With the gate removed, 5 seekpath tests failed on the `ubuntu 3.14, optional` and `macos 3.13, optional` legs with `TypeError: 'NoneType' object is not callable` at `kpath.py:941`. This is exactly the "runtime probe alone doesn't catch all the breakage" phenomenon behind 3643cdb4d — but the root cause turns out to be a test-isolation bug elsewhere in the suite, not a platform/Python-version incompatibility:

`test_err_msg_on_seekpath_not_installed` (`tests/io/pwmat/test_inputs.py`) reloads `pymatgen.symmetry.kpath` while seekpath is shadowed out of `sys.modules`. The reload executes the module's `try/except ImportError` fallback, leaving `get_path = None` in its namespace. When `patch.dict` exits, `sys.modules` is restored but the module namespace stays poisoned — nothing ever reloads it back. Every later `KPathSeek` construction then fails at `kpath.py:941` (the only call site of the module-level `get_path`).

Ordering explains the original symptoms:

- tests run alphabetically (`io/pwmat` < `symmetry`), so the polluter runs before the victims;
- the probes in this PR are evaluated at collection time, before any test executes, so they see a healthy seekpath and pass — a probe at collection time cannot catch run-time pollution;
- the suite runs under pytest-xdist (`-n auto` in `addopts`, and `[gw1]` in the failure log), so pollution only hits workers that also ran the pwmat tests — which made the original failure look platform-specific and intermittent.

So the gate from 3643cdb4d was suppressing a real, current failure — but the failure was never about Windows or Python ≥ 3.13.

The second commit of this PR restores the real modules after the simulation (in a `finally`, so a failing assertion inside the block cannot skip the restore). Verification:

- poisoned trio `tests/io/pwmat/test_inputs.py + tests/symmetry/test_kpath_hin.py + tests/symmetry/test_kpaths.py` in a single process, Windows py3.13 + seekpath 2.2.1 / spglib 2.7.0: **5 failed → 26 passed**, failure set identical to the CI failures (`test_kpath_generation`, `test_kpath_acentered`, `test_kpath_hinuma`, `test_kpath_all_combines_three`, `test_kpath_generation_across_lattices`);
- `tests/io/pwmat` alone: 19 passed — the simulation test itself still exercises the intended error path;
- `tests/symmetry`: 100 passed on py3.13, same on py3.14; without the optional extra the probe still skips cleanly; ruff clean.

Yi-111-a:
Follow-up with the decisive data: the fix was verified on real Linux and macOS runners.

I re-ran the full test matrix on my fork (same workflow, same matrix) with the two commits of this PR:

| job | before (`edebc37e9`, probe only) | after (`9d5d3267d`, probe + pollution fix) |
|---|---|---|
| ubuntu 3.14, highest, optional | 5 failed / 2940 passed | **0 failed / 2945 passed**, 47 skipped |
| macos 3.13, lowest-direct, optional | 5 failed | **0 failed / 2897 passed**, 86 skipped |
| ubuntu 3.11, lowest-direct | ✓ | ✓ |
| windows 3.13, lowest-direct | ✓ | ✓ |

The +5 passed on ubuntu are exactly the previously-failing seekpath tests; skipped counts are unchanged, so nothing else moved. Run: https://github.com/Yi-111-a/pymatgen-core/actions/runs/34614064177

That closes the last unverified surface — the gate can be removed on all platforms, not just the ones I could test locally.

shyuep:

The core finding is solid and well-evidenced: the `sys.platform` / `sys.version_info` gate plus the 2026-05-19 matrix change means the seekpath k-path tests run on **zero** CI legs today, and the upstream cause (seekpath#118, fixed in 2.2.1) is already pinned away in `uv.lock`. The `test_err_msg_on_seekpath_not_installed` module-reload leak you found while verifying is a genuine second bug and the `try/finally` restore is the right shape.

Four things I would change before merge:

1. **`except Exception: return False` converts real regressions into silent skips.** That is the same failure mode this PR exists to fix, one level down: if a future change breaks `KPathSeek` outright, the probe swallows it and the whole module skips with "seekpath not usable in this environment" instead of failing. Narrow it to the signature you actually documented — `except TypeError` — or at minimum re-raise anything that does not match `'NoneType' object is not callable`, and put the caught exception text into the skip reason so the log says *why*.

2. **The probe is duplicated across two files with a "kept in sync" comment.** That comment is a prediction of drift. Put `_seekpath_works()` in `pymatgen.util.testing` (or a `tests/symmetry/conftest.py` fixture) and import it in both. Seven `KPathSeek(...).get_kpoints()` calls at import time, twice, also runs at *collection* time in every worker — worth an `@functools.cache` on the helper.

3. **Reload-based module surgery is still fragile even with the `finally`.** `importlib.reload(pymatgen.symmetry.kpath)` rebinds the module's classes, so any test module that did `from pymatgen.symmetry.kpath import KPathSeek` at import time keeps a stale class object for the rest of the session — `isinstance` checks against the reloaded class will silently fail. Since the thing under test is just the `get_path is None` fallback, `monkeypatch.setattr(pymatgen.symmetry.kpath, "get_path", None)` (plus the same on `pymatgen.io.pwmat.inputs` if it re-exports) gets the same coverage with no reload and no `finally`.

4. **CI has never run on this branch** (first-time-contributor approval pending), so the only evidence is your fork's matrix. The fork results are convincing, but this should not merge until the real matrix is green — especially the `windows-latest 3.13` leg, which under this PR goes from "always skipped" to "runs for the first time".

Nothing here is a correctness objection to the change itself; 1 is the one I would not merge without.

Yi-111-a:
Thanks for the review — (1) in particular was a fair hit, and I'd missed it. Addressed in d627f5c1..4a7e218a.

**(1) the silent skip.** Fixed. The probe now tolerates only the two documented unavailability signatures and re-raises everything else, and the exception text goes into the skip reason:

```python
except (RuntimeError, TypeError) as exc:
    # seekpath missing: raised by the @requires guard on KPathSeek.__init__
    missing = isinstance(exc, RuntimeError) and "SeeK-path needs to be installed" in str(exc)
    # spglib returned no dataset, so kpath.get_path was left as None
    no_dataset = isinstance(exc, TypeError) and "'NoneType' object is not callable" in str(exc)
    if not (missing or no_dataset):
        raise
    return f"seekpath unusable in this environment: {type(exc).__name__}: {exc}"
```

One correction to the suggested fix, in case it matters later: narrowing to `except TypeError` alone is not enough. When seekpath simply isn't installed, `KPathSeek(...)` raises **RuntimeError** from the `@requires` guard, not TypeError — so those legs would go from "skip" to "error". Both signatures are needed.

**(2) duplication.** Agreed, the "kept in sync" comment was a promise to drift. `seekpath_unusable_reason()` now lives once in `pymatgen.util.testing` and both modules import it, with `@functools.cache` so the seven `KPathSeek` constructions run once per session.

**(3) the reloads.** I tried this and it does not work — the reload is load-bearing. `KPathSeek.__init__` is guarded by `@requires(get_path is not None, ...)`, and monty's `requires.__init__(self, condition: bool, ...)` captures that condition as a bool **at decoration time**. Once `kpath` has been imported with seekpath importable, the guard is permanently satisfied for that class object, so blanking `get_path` afterwards only changes what `kpath.py:941` calls. Measured on pymatgen-core 2026.8.30 / seekpath 2.2.1 / spglib 2.7.0:

```
>>> kpath.get_path = None
>>> KPathSeek(struct).get_kpoints()
TypeError: 'NoneType' object is not callable
```

The test asserts `RuntimeError` "SeeK-path needs to be installed", so it would fail. Re-executing the module is what re-evaluates the guard. (I've put this in a comment at the test so it doesn't get "simplified" later.)

On the stale-class-object concern: agreed it's a real hazard in general, but it doesn't bite here — `importlib.reload` re-executes into the *same* module namespace, so a class captured before the reload still resolves `get_path` through `kpath.__dict__` and heals once the restore runs. Rather than leave that implicit I made it an assertion, so if the restore ever regresses the failure lands in this test instead of five unrelated ones:

```python
assert pymatgen.symmetry.kpath.get_path is not None, "kpath.get_path was not restored"
```

**(4) CI.** Understood; not asking for merge before the real matrix is green. For what it's worth I've now been able to install the stack locally, so the last unverified surface is smaller. On Windows, Python 3.13, pymatgen-core 2026.8.30, seekpath 2.2.1, spglib 2.7.0, single process:

| run | result |
|---|---|
| `tests/io/pwmat/test_inputs.py` + `tests/symmetry/test_kpath_hin.py` + `tests/symmetry/test_kpaths.py` | **26 passed** |
| same trio with the `finally` restore removed | **6 failed** — the new assertion, plus the same 5 seekpath tests that failed on CI |

`ruff check` and `ruff format` are clean on all four files.

shyuep:

Re-review after d627f5c1..4a7e218a. Points 1 and 2 are resolved well — the probe now re-raises anything outside the two documented signatures and puts the exception text in the skip reason, and `seekpath_unusable_reason()` lives in one place with `@functools.cache`. Your correction on `RuntimeError` (the `@requires` guard, not `TypeError`) is right and my suggestion was wrong.

I also withdraw point 3. monty's `requires.__init__` takes `condition: bool`, so the guard is fixed at decoration time and `monkeypatch.setattr(kpath, "get_path", None)` genuinely cannot reproduce the `RuntimeError` — re-executing the module is the only way to re-evaluate it. The comment you added at the test is the right defense against a future "simplification".

One new blocker, introduced by the fix rather than present before:

**`assert pymatgen.symmetry.kpath.get_path is not None` will fail on the two legs that do not install the `optional` extra.** `seekpath>=2.0.1` is declared only in `optional`, and CI installs `$WHEEL_FILE[${{matrix.config.extras}}]` — so `windows-latest / 3.13 / lowest-direct` and `ubuntu-latest / 3.11 / lowest-direct` run without seekpath at all. `tests/io/pwmat/test_inputs.py` has no module-level seekpath gate, so on those legs the restore reload re-executes `kpath` with seekpath genuinely absent, `get_path` is legitimately `None`, and the new assertion turns a currently-passing test into a failure. The assertion is checking "seekpath is installed" when it means "the reload was undone".

Capture the value first and assert restoration to it:

```python
get_path_before = pymatgen.symmetry.kpath.get_path
...
assert pymatgen.symmetry.kpath.get_path is get_path_before, "kpath.get_path was not restored"
```

That is `None is None` on the no-seekpath legs and a real check everywhere else. Gating the assertion on `find_spec("seekpath")` works too but says less.

Two minor things:

- `seekpath_unusable_reason()` is now public API in a shipped module (`pymatgen.util.testing`), not test-only code. Fine if intended — `MatSciTest` is already there — just flagging that it is now covered by the deprecation policy.
- The trigonal entry in the probe's lattice tuple is commented `# rhombohedral` but is `Lattice.hexagonal(2, 95)`. Same comment appears in #140; the lattice is correct, the label is not.

The probe runs at collection time and `test_inputs.py`'s reload runs at execution time, so the cache cannot be poisoned by it under normal pytest ordering — worth keeping in mind if collection ever becomes lazy, but not something to design around now.

Nothing else blocks. Please get the real matrix green, especially `windows-latest / 3.13`, before merge.

Yi-111-a:
Both fixed in d063dfe:

1. **Restore assertion.** You're right that `is not None` was checking "seekpath is installed" when it means "the reload was undone". It now captures `get_path_before = pymatgen.symmetry.kpath.get_path` before the reload and asserts `pymatgen.symmetry.kpath.get_path is get_path_before` after the restore — `None is None` on the no-`optional` legs, a real check everywhere else. Applied exactly as suggested.

2. **Comment label.** The probe's `Lattice.hexagonal(2, 95)` entry is now labeled `# trigonal (hexagonal axes)`. I checked `tests/symmetry/test_structure.py` on `main` (via #140) and it has no such mislabeled comment anymore, so nothing to follow up there.

On the public-API flag for `seekpath_unusable_reason()`: it is intended — the probe is useful to any code that needs a seekpath availability check with a reason, and it sits next to `MatSciTest` in `pymatgen.util.testing`. Understood that it's now covered by the deprecation policy.

Locally verified (Windows, Python 3.13, pymatgen-core 2026.8.30, seekpath 2.2.1, spglib 2.7.0): the three test modules give 26 passed; `ruff check` / `ruff format` clean on both touched files. Watching the CI matrix, will report when `windows-latest / 3.13` is green.

shyuep:

Re-review of `d063dfe`. Both previous points are properly resolved: the restore assertion now compares against the captured pre-reload value (so it is meaningful on legs without seekpath), and the probe comment is fixed. The direction — probe the environment instead of gating on `sys.platform` / `sys.version_info` — is right, and the narrow exception filter that re-raises anything other than the two known signatures is the correct conservative choice.

One substantive issue remains:

1. **The seven-lattice probe does not exercise seven crystal systems.** The three sites are `K`, `La`, `Ti` at general positions with no symmetry relating them, so spglib will resolve *P*1 regardless of the lattice metric, and seekpath will take the triclinic branch on all seven. The probe therefore costs 7× the work for 1× the coverage, and the docstring's "one lattice per crystal system exercised by the tests" is not what the code does. Either (a) use structures that actually realize each system — rock salt / rutile / wurtzite / etc., which is what the downstream `test_kpath_generation_across_lattices` really depends on — or (b) drop to a single cell and say plainly that the probe only establishes that `KPathSeek` is callable, not that every Bravais path works. (b) is fine by me and much cheaper; (a) is what the current docstring promises.

Minor:

2. `@functools.cache` means the probe result is frozen for the session. That is what you want given the `test_kpath_hin` / `test_kpaths` module-level calls run at collection time, i.e. before `test_err_msg_on_seekpath_not_installed` can poison `kpath.get_path` — but the correctness of the whole arrangement rests on that ordering. Worth one sentence in the docstring saying the cache is deliberately populated at collection time for exactly this reason.
3. `seekpath_unusable_reason` now lands in the public surface of `pymatgen.util.testing`. Fine for a testing module, but if that module has an `__all__`, add it there so it is intentional rather than incidental.

CI green across all five legs.

Yi-111-a:
Addressed the re-review in 2d21258:

1. **Single-cell probe.** Dropped the seven-lattice loop (it always resolved *P*1 with general-position sites) to one cubic cell. Docstring now says plainly that the probe only establishes `KPathSeek` is callable, not that every Bravais path works — that coverage stays in the seekpath test modules.

2. **Cache note.** Documented that `@functools.cache` is deliberately populated at collection time so a later simulated absence of seekpath cannot flip the skip decision.

3. **`__all__`.** `pymatgen.util.testing` has no `__all__`, so nothing to add.

CI matrix watching.

shyuep:

Good change — the analysis that the platform/version gate made these tests dead on *every* CI leg is convincing, and replacing it with a runtime capability probe is the right fix. The `try/finally` reload restoration in `test_err_msg_on_seekpath_not_installed` is a genuine fix for cross-module test pollution. CI green.

**Issues**

1. **Docstring contradicts the PR description.** The PR body says "an end-to-end probe across seven crystal systems"; the implementation probes a single P1 cell, and the docstring correctly says so. Please reconcile the description — as written a reviewer would expect seven probes.

2. **Probe cell has out-of-cell coordinates.** `[[0.345, 5, 0.77298], [0.1345, 5.1, 0.77298], [0.7, 0.8, 0.9]]` — fractional `y = 5` and `5.1` on a 2 Å cubic cell. It works, but two sites differing by 0.1 in an unwrapped coordinate is an odd way to write "three general positions". Use in-cell values (and `to_unit_cell=True` if you want it guaranteed).

3. **Exception allowlist is string-matched.** `"SeeK-path needs to be installed"` and `"'NoneType' object is not callable"` will silently stop matching if monty's `@requires` message or CPython's `TypeError` wording changes, converting a skip into a hard suite failure. Deliberate per the docstring, but consider checking `kpath.get_path is None` directly for the second case instead of the message text.

4. **`@functools.cache` + import-order dependence.** The comment about callers "deliberately populating the cache before any test can poison `kpath.get_path`" encodes a collection-order contract between `tests/io/pwmat` and `tests/symmetry`. With `pytest -p xdist` or `--random-order` that contract is not guaranteed. Now that the pwmat test restores the module in `finally`, is the cache still load-bearing? If not, dropping `@functools.cache` (or exposing a `cache_clear` in a fixture) removes the hidden coupling.

5. `seekpath_unusable_reason` lands in `pymatgen.util.testing`, which ships in the wheel — so it is public API. Fine if intended; a leading underscore would signal otherwise.

**Nit:** the comment volume is well above the surrounding file's norm — several blocks restate the same reload rationale three times. Condensing would help future readers.

Yi-111-a:
Addressed the 2026-09-14 automated re-review in 40ceb92:

1. **PR description reconciled.** Body no longer claims a seven-crystal-system probe; it matches the implementation (single cubic P1 cell that only establishes `KPathSeek` is callable).

2. **In-cell probe coords.** Replaced `y=5` / `y=5.1` with `0.0` / `0.1` and set `to_unit_cell=True`.

3. **Exception allowlist.** Dropped string-matching of monty/CPython messages. Both known unavailability paths leave `kpath.get_path is None`; anything else is re-raised.

4. **`@functools.cache` removed.** With the `finally` restore in `test_err_msg_on_seekpath_not_installed`, the collection-order contract is no longer load-bearing. One cubic-cell probe is cheap enough to run per caller.

5. **Public API.** Leaving `seekpath_unusable_reason` un-underscored intentionally (same rationale as before — useful next to `MatSciTest`).

Also condensed the reload comments in `test_inputs.py` (nit).

shyuep:

All five items from the 2026-09-14 review are addressed in `40ceb92`, and CI is green on the head commit (lint + all four test legs). The `get_path is None` check in place of message matching, the in-cell probe coords, and dropping `@functools.cache` now that `test_err_msg_on_seekpath_not_installed` restores in `finally` are all the right calls. Point 5 (public API) — accepting your rationale.

One residual issue worth addressing before merge:

**Blast radius of the deliberate re-raise.** `seekpath_unusable_reason()` catches only `RuntimeError`/`TypeError`, and in `test_kpaths.py` it is called at module scope with no `importorskip` guard. If seekpath is installed but broken in a way that surfaces as anything else — a `ValueError` or spglib error out of `get_path`, say — the exception propagates during collection and takes down the *whole file*, including `test_setyawan_curtarolo_attaches_prim_attrs` and the other tests that never touch seekpath. The old `_seekpath_works()` caught bare `Exception`, so this is a behaviour change in the failure mode, not just in the skip logic.

The intent ("a real `KPathSeek` regression fails the suite instead of silently skipping") is right, but a collection error is a blunt way to express it. Two options that keep the intent without the collateral: evaluate the probe lazily in a fixture so only the seekpath-marked tests error, or keep the module-scope call but let it return a reason for any exception while emitting a warning on the unexpected ones.

**Nits**

- The probe now runs once per importing module (two spglib symmetry analyses per session). Negligible cost, but if `KPathSeek` ever warns under a `filterwarnings = error` config, that now fires at collection rather than inside a test.
- `assert pymatgen.symmetry.kpath.get_path is get_path_before` covers the `kpath` restore but not `pymatgen.io.pwmat.inputs`; a cheap `GenKpt` identity check would close the loop on both reloads.

Otherwise this is a clean fix for a gate that was suppressing coverage on every CI leg. Good writeup of the root cause and timeline.

Yi-111-a:
Addressed the residual items in 60b8726:

1. **Probe blast radius → lazy fixture.** `seekpath_unusable_reason()` still re-raises anything outside the two documented signatures, but it's now evaluated by a module-scoped `seekpath_usable` fixture in `tests/symmetry/conftest.py` instead of at module scope. An unexpected probe error now errors only the seekpath-dependent tests (via `usefixtures` in `test_kpaths.py`, `pytestmark` in `test_kpath_hin.py`) rather than aborting collection of the whole file — `test_setyawan_curtarolo_attaches_prim_attrs` et al. are unaffected. Module scope keeps it to one probe per file, so the cost stays at two spglib analyses per session.

2. **Second reload covered.** `test_err_msg_on_seekpath_not_installed` now also asserts `pymatgen.io.pwmat.inputs.KPathSeek is pymatgen.symmetry.kpath.KPathSeek` — if the `inputs` restore reload regresses, its `KPathSeek` stays bound to the poisoned reload's class object and this fails. (`GenKpt` identity can't serve here — it's defined in `inputs` itself, so a reload always recreates it; the imported `KPathSeek` binding is the observable that tracks the reload.)

On the `filterwarnings = error` nit: the probe now runs inside a fixture (test setup), so any future `KPathSeek` warning lands on the seekpath tests rather than at collection — consistent with the rest.

Verified locally (Python 3.13.15, seekpath 2.2.1, spglib 2.7.0): `tests/symmetry` + `tests/io/pwmat` → 100 + 19 passed with seekpath installed; without seekpath the fixture skips the four marked tests and `importorskip` handles the rest; `ruff check`/`ruff format --check` clean.

shyuep:

Re-review of `60b8726`, the only commit since the 2026-09-15 pass. CI is green on all five legs.

The residual issue from the last review is resolved: the probe now runs lazily through a module-scoped `seekpath_usable` fixture in `tests/symmetry/conftest.py`, applied via `pytestmark` in `test_kpath_hin.py` and per-test `usefixtures` in `test_kpaths.py`. An unexpected exception out of `KPathSeek` now errors only the seekpath-dependent tests rather than failing collection of the whole module, and `test_setyawan_curtarolo_attaches_prim_attrs` is unaffected. Both earlier nits are also addressed (one probe per module; `pwmat.inputs.KPathSeek is kpath.KPathSeek` closes the restore check on both reloads).

No further comments from this pass — ready for a human maintainer to merge.

shyuep:
Thanks.

Issue #139, opened 2026-09-11
get_symmetry_dataset(backend="moyopy") disagrees with backend="spglib" on "international" (all 13 monoclinic groups) and "std_origin_shift" (60 groups)


Yi-111-a:
# Issue draft: pymatgen-core — cross-backend inconsistency in `Structure.get_symmetry_dataset`

**Title:** `get_symmetry_dataset(backend="moyopy") disagrees with backend="spglib" on "international" (all 13 monoclinic groups) and "std_origin_shift" (60 groups)`

## Summary

Sweeping all 230 spacegroups and comparing the dict output of the two symmetry backends
(`return_raw_dataset=False`, the default), the fields that describe symmetry semantics
(`number`, `orbits`, `wyckoffs`, `site_symmetry_symbols`) agree perfectly. Two fields diverge:

1. **`international` differs for all 13 monoclinic groups (3–15).**
   The spglib path returns spglib's short Hermann–Mauguin symbol (`P2`, `C2/c`), while the
   moyopy path returns `SpaceGroup.from_int_number(dataset.number).symbol`, which pymatgen
   canonicalizes to the *full* HM symbol (`P121`, `C12/c1`). `MoyoDataset` has no
   `international` attribute, so the moyopy path has to derive the symbol — but it derives it
   through a different convention than the spglib path exposes. Any user code that compares
   symbols across backends (or switches backends) sees different strings for the same structure.

2. **`std_origin_shift` differs for 60 groups** (a further 71 diffs are float noise < 1e-6).
   Example, SG 9 (Cc): spglib `[0, 0, 0]` vs moyopy `[-0.5, 0.5, 0.5]`; SG 48: `[-0.25, -0.25, -0.25]`.
   This is consistent with the two backends standardizing to different origin choices, and it
   silently changes the meaning of the returned standardized cell: anyone using
   `std_origin_shift` to map coordinates into the standardized cell gets different results
   depending on the backend.

## Repro

```python
# full sweep script attached (moyopy_sweep.py): builds one structure per spacegroup 1-230
# via Structure.from_spacegroup and compares get_symmetry_dataset(backend="moyopy")
# against backend="spglib", default dict path.
# Result (moyopy 0.18.0, spglib 2.7.0, pymatgen-core 2026.8.30, Windows, py3.13):
#   230/230 completed, 0 errors
#   mismatches: international=13 (SG 3-15), std_origin_shift=131 (60 real + 71 float noise)
[...]

PR #140, opened 2026-09-11
tests: cover the moyopy symmetry backend and pin a working floor

Yi-111-a:
## Summary

`Structure.get_symmetry_dataset(backend="moyopy")` is a public API, but it has no CI coverage: no matrix leg installs `moyopy` (the `optional` extra contains netcdf4/ase/beautifulsoup4/f90nml/h5py/hiphive/phonopy/seekpath only), so the moyopy branch of `test_get_symmetry_dataset` is `importorskip`-gated out of every job. On top of that, the existing test only exercises `return_raw_dataset=True` — the default dict-conversion path has never been executed anywhere.

Found while investigating the cross-backend divergences documented in #139.

## Changes

- Add `moyopy` to the `optional` extra so the macos 3.13 and ubuntu 3.14 legs install it. Plain `moyopy`, not `moyopy[interface]` — the interface extra pulls the full `pymatgen` package, which would be circular for this repo.
- New parametrized test sweeping all 230 spacegroups, asserting both backends agree on the backend-independent semantics (`number`, `wyckoffs`, `site_symmetry_symbols`, `orbits`).
- Floor `moyopy>=0.17`, determined empirically: with 0.3.0 four spacegroups get wyckoff letters that disagree with spglib, and 0.16.0 still disagrees on sg 42/69; 0.17.0 and 0.18.0 agree on all 230. The `symmetry` extra still declares `moyopy[interface]>=0.3` — raising that floor looks warranted too, but is left out of this PR.
- Known intentional divergences (see #139) are documented in the test and deliberately not asserted: `international` (short vs full HM symbol), `std_origin_shift` and `hall_number` (different origin-choice conventions, affecting 60 and 24 groups respectively).

## Verification

- Windows py3.13, moyopy 0.18.0: 230/230 spacegroups agree; full `tests/core/test_structure.py` = 396 passed / 10 skipped.
- Same sweep on moyopy 0.17.0: 230/230 agree; on 0.16.0 sg 42/69 fail; on 0.3.0 four groups fail — hence the floor.
- Without moyopy installed, all 230 parametrized cases skip cleanly (checked on a no-optional environment).


Yi-111-a:
Verified on real runners: I re-ran the full matrix on my fork (same workflow, `optional` legs now install moyopy via this PR) — all four jobs green:

| job | moyopy | result |
|---|---|---|
| ubuntu 3.14, highest, optional | 0.18.0 | **3170 passed** / 51 skipped (previously 2945 — the ~225 delta is the new sweep actually running) |
| macos 3.13, lowest-direct, optional | 0.18.0 | **3119 passed** / 91 skipped |
| ubuntu 3.11, lowest-direct (no extras) | — | ✓ (sweep skips cleanly) |
| windows 3.13, lowest-direct (no extras) | — | ✓ 2754 passed / 444 skipped — identical passed count to the pre-PR baseline, +230 skips = the sweep skipping cleanly |

Run: https://github.com/Yi-111-a/pymatgen-core/actions/runs/34620241248

shyuep:

Good catch — `backend="moyopy"` is public API with no CI coverage at all, and the default `return_raw_dataset=False` conversion path has never executed anywhere. The lattice-per-crystal-system mapping (1-2 triclinic, 3-15 monoclinic, 16-74 ortho, 75-142 tetragonal, 143-167 trigonal, 168-194 hexagonal, 195+ cubic) is correct, and excluding `international` / `std_origin_shift` / `hall_number` per #139 is the right call.

Four points:

1. **The user-facing floor is the one that matters, and this PR does not raise it.** You established empirically that moyopy < 0.17 produces Wyckoff letters that disagree with spglib on sg 42/69 (and four more at 0.3), then raised the floor only in the `optional` (test) extra while `symmetry` still declares `moyopy[interface]>=0.3`. That leaves users of `pip install pymatgen[symmetry]` resolving to a version you have positive evidence is wrong. It is a one-line change in the same file with the evidence already in this PR — please fold it in rather than deferring.

2. **230 parametrized cases is a lot of CI wall-clock for what is really one property.** Each case builds a `Structure.from_spacegroup` with 3 species on general positions (up to 192-fold multiplicity in the cubic groups) and runs two symmetry analyses. Consider `@pytest.mark.parametrize("sg", ...)` over a representative subset plus the known-divergent groups (42, 69), or keeping the full sweep behind a marker. If you keep all 230, hoist `pytest.importorskip("moyopy")` to module level or a `pytest.mark.skipif` so the no-optional legs report one skip instead of 230.

3. **`lowest-direct` resolution will pin exactly `moyopy==0.17`** on the macOS leg. Your verification covered 0.17.0 and 0.18.0, so that is fine — just noting the floor is now load-bearing for CI, not just advisory.

4. **Test placement.** `test_symmetry_dataset_backends_agree_on_semantics` is a cross-backend contract test, not a `Structure` test; `tests/symmetry/` would be easier to find. Minor.

CI has not run on this branch yet (first-time-contributor approval) — the added dependency install on two legs should be confirmed green before merge.

Yi-111-a:
Thanks for the review — both points addressed in 27c11b8 (force-pushed after I accidentally landed a wrong-file commit on this branch; the final state contains only the two files listed).

**1. User-facing floor raised.** `symmetry` extra now declares `moyopy[interface]>=0.17`, same as the `optional` extra, with a comment noting the reason. So `pip install pymatgen[symmetry]` can no longer resolve to a version where Wyckoff/orbit assignments disagree with spglib (sg 42/69 and four more at 0.3).

**2. Collection-time gating.** The 230-case sweep now checks `find_spec("moyopy")` once (module level, `skipif` marker) instead of calling `pytest.importorskip()` inside the test body 230 times. Measured on a leg without the optional dep: 230 skipped in 1.42 s.

On cutting the sweep down: I measured before touching it — the full 230-spacegroup sweep costs **7.4 s** on this machine (Windows, Python 3.13, moyopy 0.18.0). That is small enough that I would rather keep all 230: the sweep is what surfaced the 0.3-vs-0.17 divergence in the first place, and thinning it would weaken exactly the check that caught the bug. Happy to cut to a representative subset plus {42, 69} if you prefer; it is a one-line change to the parametrize.

Not done: moving the test to `tests/symmetry/`. Agreed it is a better home, but I did not want to move a test across files in the same pass as changing its behaviour — say the word and I will do it as a separate commit.

Verified locally with pymatgen 2026.8.30 + moyopy 0.18.0 + spglib 2.7.0: 230 passed with moyopy installed, 230 skipped with it hidden. `ruff check` and `ruff format` clean.

shyuep:

Re-review after 27c11b8. Points 1 and 2 are addressed correctly — the `symmetry` extra now carries the same `>=0.17` floor as `optional`, and the `find_spec` + module-level `skipif` replaces 230 `importorskip` calls with one collection-time decision. Your measurement (7.4 s for the full sweep) settles the wall-clock question; keep all 230, and the point 4 test move is fine as a separate commit.

One thing the new diff introduces:

**The two extras now declare different moyopy requirements, and the weaker one is what CI installs.** `symmetry` declares `moyopy[interface]>=0.17`; `optional` declares bare `moyopy>=0.17`. CI installs `$WHEEL_FILE[optional]` — it never installs the `symmetry` extra — and the code path under test does:

```python
import moyopy.interface
moyo_cell = moyopy.interface.MoyoAdapter.from_structure(self)
```

That works on the optional legs today only because `ase>=3.27.0` happens to be listed in the same extra. The dependency the test actually needs is incidental rather than declared: drop `ase` from `optional`, or have moyopy move what `[interface]` pulls in, and 230 tests fail with the misleading `ImportError("moyopy is not installed. Run pip install moyopy.")` from `structure.py`. Make `optional` declare `moyopy[interface]>=0.17` too — same one-line argument you accepted for the floor itself.

Related: `find_spec("moyopy")` then becomes the wrong gate for the same reason. `find_spec("moyopy.interface")` is what the test's code path requires, and it degrades to a skip rather than 230 errors if the interface extra is ever absent.

Also minor, and it recurs in #138: the trigonal branch is commented `# rhombohedral` but builds `Lattice.hexagonal(2, 95)`. The lattice is right for `from_spacegroup` on 143-167 (hexagonal setting); the comment is what is wrong.

CI has still not run on this branch. Since this PR adds a dependency install to two legs, please confirm green before merge rather than relying on the fork matrix.

shyuep:
Thanks for the improvement in coverage. When monopy was first added, it wasn't mature. But it has considerably in the past year. We may consider making it the default going forward.

PR #141, opened 2026-09-11
test-files: add MoleculeGraph reference data missing from the repo split

Yi-111-a:
## Summary

`tests/core/test_graphs.py::TestMoleculeGraph::test_construction` loads
`{TEST_FILES_DIR}/analysis/bond_dissociation/pc_frag1_mg.json` and `pc_mg.json`, but those files were never copied into this repo when the test suite was split out of `materialsproject/pymatgen` — they only exist in the `pymatgen-test-files` repository, which the main repo uses as a submodule. This repo carries its own in-tree `test-files/` directory, and it lacks the whole `analysis/bond_dissociation/` directory.

The breakage is invisible to CI because the test branch is `importorskip`-gated on openbabel, which no matrix leg installs — with openbabel available the test dies immediately on a `FileNotFoundError`.

## Changes

- Add the two reference files verbatim from `pymatgen-test-files@HEAD` (`f70fd8852`, 2026-02-20). They are compatible with the current `MoleculeGraph` schema — no regeneration needed.

## Verification

- Windows py3.13 with `openbabel-wheel` 3.1.1.23: `tests/core/test_graphs.py` = **32 passed / 1 skipped / 1 xfailed** (previously: 1 failed on `FileNotFoundError`).
- The other openbabel-gated files also pass with openbabel installed: `tests/core/test_local_env.py`, `tests/io/test_packmol.py` (76 passed alongside test_graphs in the combined run).
- Without openbabel the `importorskip` behavior is unchanged.


shyuep:

Straightforward and correct: two reference JSON files (1.8 kB and 4.2 kB) restored verbatim from `pymatgen-test-files@f70fd8852` into the in-tree `test-files/`, fixing a `FileNotFoundError` left behind by the repo split. No source changes, no risk.

Two notes:

1. **The fix is still not verifiable by CI.** `TestMoleculeGraph::test_construction` is `importorskip`-gated on openbabel, which no matrix leg installs — so this PR restores files that CI will continue to never read, and the same drift can recur silently. Either add `openbabel-wheel` to the `optional` extra (which is what you used locally) so the gate opens on the two optional legs, or add a cheap guard test that asserts the referenced `TEST_FILES_DIR` paths exist, independent of openbabel. The second is one line and catches exactly this class of split-artifact regression.

2. **Worth a sweep rather than a spot fix.** If `analysis/bond_dissociation/` went missing in the split, other directories referenced only from openbabel/enumlib/vampire-gated tests likely did too. A one-off scan of `TEST_FILES_DIR`-relative literals in `tests/` against the tree would tell you whether #141 is the whole problem or the first instance.

Neither blocks merge. Please confirm the added files are under the same license terms as the rest of `test-files/` (the directory carries its own LICENSE).

Yi-111-a:
All three notes addressed in d05ed2a:

**1. Guard test added** — `TestMoleculeGraph::test_bond_dissociation_reference_files_exist` runs on every leg, independent of openbabel, and asserts the two reference JSONs exist with a message naming the file and why CI cannot otherwise notice (`test_construction` is importorskip-gated on openbabel, which no matrix leg installs). Verified: 1 passed with the files present; and the assertion catches the pre-fix state (in-tree `test-files/` without `analysis/bond_dissociation/`).

I went with your second option (guard test) rather than adding `openbabel-wheel` to the optional extra — the guard is one method and catches this exact class of regression on every CI leg, whereas the extra only opens the gate on the legs that install it.

**2. Sweep done** — I extracted every resolvable `TEST_FILES_DIR`-relative literal across `tests/` (416, with chained f-string prefix variables resolved) and checked them against the tree. **Nothing else is missing.** The only other gaps the scan surfaced are expected:
- runtime-generated: `unk_dir/` written by `write_unks` in the WAVECAR test;
- shipped inside the package, not test-files: `lobster_basis/BASIS_PBE_54_standard.yaml` (resolved via `pymatgen.io.vasp.sets.MODULE_DIR`);
- runtime-joined list items (the `new_qchem_files/` and `qchem_energies/` entries in the qchem tests) — all present in the tree.

So #141 was the whole problem, not the first instance. Caveat, stated honestly: this is static analysis of string literals; paths built via `os.path.join` on computed names are outside its reach.

**3. License** — both files were restored verbatim from `pymatgen-test-files`, which carries its own MIT LICENSE (Copyright (c) 2026 Materialyze.AI); same license family as this repo (MIT). The directory's LICENSE governs its contents, so the in-tree copies inherit that.

The sweep script is one-off; happy to check it in as a test helper if you would rather have the sweep runnable in-repo.

shyuep:

Re-review after d05ed2a. All three prior notes are properly addressed: the guard test runs on every leg independent of openbabel, the 416-literal sweep confirms this was the whole gap rather than the first instance, and the license question is settled (both directories MIT). The choice of the guard test over adding `openbabel-wheel` to the `optional` extra is the right one — it catches the regression class on all four legs instead of two.

One blocker:

**`tests/core/test_graphs.py` was committed with CRLF line endings.** `main` is LF; the head of this branch is CRLF from line 1, so GitHub reports the file as +971/-955 — the entire 955-line file rewritten — when the actual change is one added test method. That will conflict with anything else touching this file, destroy `git blame` for the module, and may trip the `ruff format` check on a leg that normalizes line endings differently.

Please convert the file back to LF and force-push so the diff is just the new method. If your git is doing this on checkout, `git config core.autocrlf input` (or an explicit `.gitattributes` `*.py text eol=lf`) prevents a recurrence.

Everything else LGTM; happy to merge once the diff is clean and CI is green.

On the sweep script: not worth checking in as a test helper. A one-off audit that returned "nothing else missing" has no recurring signal, and the per-leg guard test you added covers the recurrence case more cheaply.

shyuep:
Thanks. Yup some things got lost in the translation.

PR #142, opened 2026-09-12
Refuse normalize_to when the target coefficient is a numerical remainder

shaneraphel:
## Summary

Fixes the blow-up reported on materialsproject/pymatgen#3983 (the module now lives in pymatgen-core).

`ComputedReaction` on an underdetermined hydrolysis set balances to

```
12 LiCuS -> Cu + Cu7S4 + 4 Li3CuS2
```

but keeps H2O / H2S in `_all_comp` with coefficients ~`6e-15`. `normalize_to(H2O)` then does `1 / 6e-15` and reports

```
1.829e+15 LiCuS + 1.153 H2S -> H2O + ...
```

with energy `4.57e11` eV (measured on pymatgen 2026.5.4 / pymatgen-core 2026.5.18).

This PR refuses `normalize_to` when `|coeff| < BalancedReaction.TOLERANCE` (`1e-6`) and when the composition is absent. Normalizing to a real participant (`LiCuS`) still works.

Data label: `pymatgen_current_ComputedReaction_normalize_to_H2O_LiCuS`. Software coefficient scaling, not a materials discovery.

## Test plan

- [x] Local reproduction of the 1e15 / 4.57e11 eV blow-up on 2026.5.4
- [x] Monkeypatched `normalize_to` raises `ValueError` matching `numerical current`
- [x] `test_normalize_to_refuses_numerical_leftover` added


shyuep:

Correct and well-scoped. Worth recording why: `BalancedReaction.__init__` already drops `abs(coeff) <= TOLERANCE` entries from `_all_comp`, so the new guard is unreachable there — it only fires on `Reaction` / `ComputedReaction`, where `_all_comp = reactants + products` is unfiltered and the underdetermined balancer can leave a ~1e-15 coefficient. That is exactly the failure in materialsproject/pymatgen#3983, so the check is in the right class. `abs(1 / c * factor)` to `abs(factor / c)` is equivalent and slightly better conditioned.

Two points:

1. **The threshold is absolute, so the guard is scale-dependent.** `TOLERANCE = 1e-6` is a composition-*amount* tolerance (it is what `almost_equals` and `_str_from_formulas` use); reusing it as an absolute coefficient floor means the guard's behaviour depends on how the reaction happens to be normalized. `rxn.normalize_to(comp, 1e-5)` scales every coefficient below 1e-6, after which a second `normalize_to` on a legitimate participant raises. A relative test matches the stated intent ("a numerical leftover, not a participant") and is normalization-invariant:

   ```python
   if abs(coeff) < self.TOLERANCE * max(abs(c) for c in self._coeffs):
   ```

2. **`normalize_to_element` has the same hole, one line down.** `current_el_amount` is `0` for an element absent from the reaction, so it raises `ZeroDivisionError` rather than anything actionable — and a trace-amount element gives the same 1e15 blow-up this PR fixes for compositions. Out of scope if you would rather keep the diff tight, but it is the same bug.

Nit: the test docstring's provenance lines (`Data label: pymatgen_current_...`, `Software coefficient blow-up, not a materials discovery`) will not mean anything to a future reader — the issue number and the before/after coefficients carry the whole story.

CI has not run on this branch (first-time-contributor approval pending).

shyuep:
I agree with the review comment 1. The numerical tolerance should be scale independent since we can potentially scale the reaction numbers up or down. E.g., if the reaction is 4000000 Li + 1000000 O2 -> 2000000 Li2O, then a coefficient of 1 somewhere might be a numerical noise.

shyuep:

Looks good. Refusing to divide by a ~1e-15 balancer leftover is the right fix, and `abs(factor / coeff)` is equivalent to the old `abs(1 / coeff * factor)` with one less rounding step. CI green.

**Points to consider**

1. **`normalize_to_element` has the same hole.** It resolves an element to a coefficient by the same route and will still divide by a leftover. Worth applying the same guard there (or factoring the check into a small helper) so the two entry points behave consistently.

2. **Hard error vs. warning.** This turns a silently-wrong result into a `ValueError`. That's defensible, but anything downstream that loops over `rxn.all_comp` and normalizes to each in turn now raises instead of producing garbage. If that pattern exists in the wild, a `warnings.warn` + no-op may be gentler; if you keep the raise, it's an API change worth a line in the release notes.

3. **Tolerance semantics.** `TOLERANCE` is a *balancing* tolerance on the least-squares residual; reusing it as an absolute floor on a coefficient conflates two different scales. It happens to work here (1e-6 vs 1e-15), but a dedicated constant would make the intent explicit and avoid coupling to a value that may be retuned for the balancer.

4. **Test assertion.** `assert max(abs(c) for c in rxn.coeffs) < 10` is a magic bound. Asserting the actual expected coefficients (or that `coeffs[all_comp.index(LiCuS)] == approx(1)` after normalizing) pins the behaviour rather than just its order of magnitude.

**Nit:** the message "The species is a numerical leftover, not a participant" asserts something the code can't know — a genuinely tiny stoichiometric coefficient would get the same message. "below the balancing tolerance and is most likely a numerical leftover" is more accurate.

shaneraphel:
tanks for review!

PR #143, opened 2026-09-12
Fix StructureGraph.__mul__ c-image on interior supercell layers

shaneraphel:
## Summary

`StructureGraph.__mul__` remaps a periodic edge into the supercell, then (when it flips `u`/`v` to keep `u < v`) negated the **pre-supercell** `to_jimage` instead of the remapped image. On the public In4Se4 layer from [pymatgen#4293](https://github.com/materialsproject/pymatgen/issues/4293) that current wrap:

- tilts the interior copy from `(0, 0, 1)` to `(1, 0, 1)`
- keeps a `(-1, 0, 1)` c-translation on a layer that now sits inside the cell

Measured on pymatgen 2026.5.4 / pymatgen-core 2026.5.18 (site-packages) and against this tree. After the one-line flip, both layers stay `(0, 0, 1)` and the interior copy has no c-wrap.

The class lives in pymatgen-core (`src/pymatgen/core/graphs.py`) after the 2026-04-07 split.

## Test plan

- [x] `tests/core/test_graphs.py::test_mul_layer_supercell_no_interior_c_wrap` (standalone; does not need MatSciTest critic2 fixtures)
- [x] Unpatched site-packages 2026.5.18 still produces current `(3, 8, (-1, 0, 1))` on the same coordinates
- [ ] CI `tests/core/test_graphs.py`


shyuep:

The one-line fix is correct. `new_to_jimage` is assigned from `v_expec_image` immediately above the swap, so it is the supercell image; negating `data["to_jimage"]` instead re-applied the *pre*-supercell wrap to a remapped edge, which is exactly the leftover c-image in materialsproject/pymatgen#4293. The corrected value also flows into the `new_periodic_images` dedup key, so that stays consistent.

One blocker and one nit:

1. **`test_mul_layer_supercell_no_interior_c_wrap` is defined twice, byte-identical.** The second definition silently shadows the first, so half the diff is dead code. Delete one.

2. **The test over-asserts on incidental structure.** `assert struct_graph.graph.number_of_edges() == 14` and `assert interior == [1, 3, 5, 7, 8, 10, 12, 14]` both depend on `MinimumDistanceNN` cutoff behaviour and on site ordering in `__mul__`'s supercell construction — neither is the property under test, and either will break this test for reasons unrelated to the bug. The `leftover_c == []` assertion is the real check; consider keeping that plus a `len(interior) == 8` sanity check and dropping the literal index list.

CI has not run on this branch (first-time-contributor approval pending).

shyuep:
Pls fix the merge conflict. Thanks.

PR #144, opened 2026-09-12
Prefer a proper rotation in Lattice.find_mapping

### LINKED ISSUE #4457: `structure.get_reduced_structure(reduction_algo="niggli")` returns a reflected structure
### Python version

Python 3.11.12

### Pymatgen version

2025.2.18

### Operating system version

MacOS 24.5.0

### Current behavior

Hi, thanks for such a great package! I am having some trouble with the Niggli reduction, and unless I am misunderstanding something I think there is a bug in the final mapping.

The issue: Running `structure.get_reduced_structure(reduction_algo="niggli")` sometimes returns a reflected version of the original system. The core issue is in the final mapping at the following lines:
https://github.com/materialsproject/pymatgen/blob/f9d9fe8e0ce09ef30cc03bcc4e9937d27afd5a6a/src/pymatgen/core/lattice.py#L1274-L1278

If you switch`skip_rotation_matrix=False` and check the determinant of the rotation matrix, you'll get -1, corresponding to a reflection matrix. 

Let me know if I'm misunderstanding something and this is an intended behavior!

Below I've provided a snippet that shows the source of the issue: the `find_mapping` method returns a reflection mapping, even when the source and target lattices are the same!

Thanks again!

### Expected Behavior

I would expect that `find_mapping` returns a rotated lattice but not a reflection. While a reflected system has the same energy, it's physical properties can be different.

### Minimal example

```Python
import ase
import ase.io
import numpy as np
from pymatgen.io.ase import AseAtomsAdaptor


def get_atoms():
    atoms = ase.Atoms(
        positions=np.array(
            [
                [0.00000000e00, 0.00000000e00, 0.00000000e00],
                [2.32160500e00, -2.55395379e-10, 2.62836300e00],
                [-2.55396948e-10, 2.32160500e00, 2.62836300e00],
                [2.32160500e00, 2.32160500e00, -6.87146875e-17],
                [2.85942205e-10, -2.85941836e-10, 2.94272573e00],
                [-4.80245259e-10, 2.32160500e00, 3.14362728e-01],
                [2.32160500e00, 2.32160500e00, 2.31400027e00],
                [-3.05465285e-11, 2.32160500e00, 4.94236327e00],
            ]
        ),
        numbers=np.array([11, 11, 65, 65, 8, 8, 8, 8]),
        cell=[
            [4.643209998612975, 0.0, 0.0],
            [-1.0215833248797727e-09, 4.643209998612975, 0.0],
            [-2.3216049982849043, -2.3216049998172785, 5.2567259980108245],
        ],
        pbc=True,
    )
    return atoms


if __name__ == "__main__":
    atoms = get_atoms()
    ase.io.write("original.png", atoms)
    ase.io.write("original.cif", atoms)

    structure = AseAtomsAdaptor.get_structure(atoms)
    # structure = structure.get_reduced_structure(reduction_algo="niggli")
    tol = 1e-5
    e = tol * structure._lattice.volume ** (1 / 3)
    # Compute a lattice mapping between it and itself
    mapping, rotation_matrix, translation_vector = structure._lattice.find_mapping(
        structure._lattice, e, skip_rotation_matrix=False
    )
    print(np.linalg.det(rotation_matrix)).    # ----> prints -1, corresponding to a reflection!
    structure_mapped = type(structure)(
        mapping,
        structure.species_and_occu,
        structure.cart_coords,
        coords_are_cartesian=True,
        to_unit_cell=True,
        site_properties=structure.site_properties,
        labels=structure.labels,
        charge=structure._charge,
    )
    atoms_mapped = AseAtomsAdaptor.get_atoms(structure_mapped)
    ase.io.write("mapped.png", atoms_mapped)
    ase.io.write("mapped.cif", atoms_mapped)
```

_No response_

shaneraphel:
## Summary

Fixes materialsproject/pymatgen#4457.

`Lattice.find_mapping` returned the first length/angle match from `find_all_mappings`. On the published Na/Tb/O cell that first match is a reflection (`det R = -1`) even though **8 proper and 8 improper** maps exist, including the identity.

Measured on pymatgen 2026.5.4 / pymatgen-core 2026.5.18: `find_mapping(self, self)` → `det(R) = -1`. After this change the same self-map returns `det(R) = +1`. An improper map is still returned when no proper map exists (true enantiomorphs). `find_all_mappings` is unchanged.

This is a lattice-mapping handedness defect, not a materials discovery. Data label: `pymatgen_2026.5.4_lattice_self_map_det_minus_one`.

## Test plan

- [x] Site-packages 2026.5.4: self-map `det(R) = -1`; 8+8 maps
- [x] Patched worktree: self-map `det(R) = +1`
- [x] `test_find_mapping` / `test_find_all_mappings` / `test_mapping_symmetry` / new `test_find_mapping_prefers_proper_rotation`


shyuep:

The diagnosis is right — `find_all_mappings` yields length/angle matches in enumeration order, and on a cell with 8 proper + 8 improper self-maps `next()` can hand back a reflection. Preferring `det > 0` and falling back to `first` is the correct shape.

Three points:

1. **The `skip_rotation_matrix` branch uses the wrong sign test.** In `find_all_mappings`, `aligned_m = scale_m @ self.matrix` and `rotation_m = solve(aligned_m, other_lattice.matrix)`, so

   ```
   det(rotation) = det(other.matrix) / (det(scale) * det(self.matrix))
   ```

   `det(scale) > 0` therefore implies `det(rotation) > 0` only when `self` and `other_lattice` have the same handedness. pymatgen does not require right-handed matrices (nothing rejects a negative-determinant `Lattice`), so the two branches can disagree on exactly the lattices where the caller most needs them to agree. The scale-invariant test is `det(scale_m) * det(self.matrix) * det(other_lattice.matrix) > 0`; that is cheap here since both determinants are already available.

2. **Worst-case cost changes from first-hit to full enumeration.** When no proper map exists, the loop now drains the entire generator before returning `first`. That is unavoidable given the goal, but `find_all_mappings` is the expensive part of `StructureMatcher`-adjacent code paths, so it is worth saying so in the docstring — as written, the new text reads as a pure semantics note.

3. **Downstream callers may now receive a different (equally valid) mapping.** `find_mapping` feeds interface/slab construction and symmetry-reduction code. This is the intended improvement, not an objection, but it is a silent behavioral change for anyone who had implicitly pinned to the old first-hit result.

Nits: the new paragraph is indented inside the `Returns:` block, so it renders as part of the return description — it belongs at docstring-body level. The test docstring carries provenance noise (`Site-packages pymatgen 2026.5.4`, `Data label pymatgen_2026.5.4_lattice_self_map_det_minus_one`, `Not a materials discovery`) that will not mean anything to a future reader; the issue reference alone is enough.

CI has not run on this branch (first-time-contributor approval pending) — please get the matrix green before merge.

shyuep:
Thanks for the fixes. I have personally reviewed the code and it looks great.

PR #55, opened 2026-05-17
Remove POTCAR cryptographic hash validation

shyuep:
## Summary

- Drop the MD5/SHA256 hash-check machinery on `PotcarSingle` and the two JSON lookup tables that backed it (~500 KB no longer shipped in the wheel).
- Keep the summary-stats validator (`PotcarSingle.is_valid` + `potcar-summary-stats.json.bz2`) — it's the intended replacement and continues to gate the `UnknownPotcarWarning` in `__init__`.
- Preserve `PotcarSingle.spec()` dict shape by retaining the `"hash"` key with value `None` (atomate / FireWorks DB rows / `Vasprun.potcar_spec` consumers don't `KeyError`).

## Motivation

The hash database is unmaintained and produces false-positive warnings on legitimate newer POTCARs. References from the parent `materialsproject/pymatgen` repo:

- materialsproject/pymatgen#4570 — *Remove POTCAR hash check* (the direct ask): _"if the warnings cannot be trusted, then they are no use at all. This part of the code should be excised altogether."_
- materialsproject/pymatgen#4467 — *`UnknownPotcarWarning` occurs for valid POTCARs due to mismatched line endings*: pristine PBE_64 La / Yb POTCARs trigger the warning unnecessarily.
- materialsproject/pymatgen#3956 — *Adding PBE_64 support: Clarification on hash types used in POTCAR hash files*: adding new POTCAR releases is blocked on regenerating the two JSON tables, which nobody does.

## Removed

- `PotcarSingle.{hash_sha256_from_file, sha256_computed_file_hash, md5_computed_file_hash, md5_header_hash}` properties
- `PotcarSingle.verify_potcar()`
- `PotcarSingle.identify_potcar_hash_based()` (incl. its 90-line `mapping_dict`)
- Module-level `PYMATGEN_POTCAR_HASHES`, `VASP_POTCAR_HASHES`
- Data files `vasp_potcar_pymatgen_hashes.json` (~217 KB) and `vasp_potcar_file_hashes.json` (~276 KB)
- `import hashlib`, `from hashlib import sha256`
- Six hash-specific test methods in `tests/io/vasp/test_inputs.py`

## Kept

- `PotcarSingle.is_valid`, `_load_potcar_summary_stats`, `_gen_potcar_summary_stats`
- `potcar-summary-stats.json.bz2`
- `UnknownPotcarWarning` and the `__init__` warning block (consults `is_valid`, not hashes)
[...]

PR #145, opened 2026-09-13
Fix PotcarSingle IndexError for POTCAR data with CRLF line endings

Yi-111-a:
Closes the IndexError variant of materialsproject/pymatgen#4467.

`PotcarSingle.is_valid` (invoked during `__init__`) split the POTCAR body on the hardcoded delimiter `"END of PSCTR-controll parameters\n"`. For POTCAR text that still carries `\r\n` — read in binary mode and decoded, or passed through a transport that does not normalize newlines — the delimiter never matches and the `[1]` index raises `IndexError` instead of parsing. The `from_file` path is unaffected because `zopen` text mode normalizes newlines, which is why this only surfaces for the string constructor.

Fix: split on `END of PSCTR-controll parameters\r?\n`. Behaviour for LF data is unchanged; CRLF data now parses to identical `keywords` and `_summary_stats` (verified on the fake `POTCAR.Mn_pv.gz` fixture).

Adds `test_parse_crlf_data` (fails on `main`, passes here). Locally verified on Windows, Python 3.13, pymatgen-core 2026.8.30: `tests/io/vasp/test_inputs.py` 96 passed; `ruff check` and `ruff format` clean on both touched files.

The false-positive-warning aspects of #4467 were already addressed by #55 and #124.


shyuep:

LGTM. Minimal, correctly scoped, and the regression test is the right one.

I checked whether the rest of `PotcarSingle` is CRLF-safe, since a one-call-site fix would be a trap otherwise — it is: `self.header` uses `.strip()`, every string-valued keyword goes through `str.strip` in `parse_functions` (numeric parsers tolerate the stray `\r` as whitespace), the `Atomic configuration` / `Description` / `RRKJ` blocks use `.splitlines()`, and the body loop's `row.split()` discards the trailing `\r`. Your assertions on `keywords`, `_summary_stats` and `is_valid` cover exactly those paths, so full-text normalization at ingest is not needed — and would have the side effect of changing `self.data` relative to the bytes on disk, so keeping the fix local is the better call.

Two optional follow-ups, neither blocking:

1. `re.split(...)[1]` still raises a bare `IndexError` when the delimiter is absent entirely (truncated or non-VASP file) — same failure mode as the bug being fixed, just a different cause. A `ValueError` naming the missing `END of PSCTR-controll parameters` marker would be a much better diagnostic.
2. The pattern does not cover lone-`\r` (classic Mac) line endings. Almost certainly not worth handling, but `\r?\n?` vs `[\r\n]+` is a one-character decision if you want it.

CI green across all five legs.

shyuep:
Thanks!

PR #146, opened 2026-09-13
Fix C2v when the C2 axis is not the unique inertia axis

shaneraphel:
## Summary
- `PointGroupAnalyzer` still reports **Cs** for a planar C2v molecule whose C2 is a principal axis but **not** the unique inertia axis (materialsproject/pymatgen#4596, author-corrected 4-atom example).
- After PR #134 the identity (360°) is no longer recorded as a rotation, and `_check_perpendicular_r2_axis` always runs. That is not enough here: the unique axis is \(y\), the smallest off-axis distance shell is a singleton, and the wing pair that generates C2(\(z\)) is never tested.
- This PR calls the existing `_check_R2_axes_asym` **only when `rot_sym` is still empty** after those two searches. It does not invent a new helper.

## Probe (pymatgen-core `25c48ed` + this change)
| molecule | before (0.01 / 0.1 / 0.3) | after |
|---|---|---|
| author-corrected 4-atom C2v | Cs / Cs / Cs | **C2v / C2v / C2v** |
| chemenv square-face capped trigonal prism | Cs / C2v / C2v | **C2v / C2v / C2v** |
| water | C2v | C2v |
| original false example (true D2d) | D2d | D2d |
| NH3-like C3v | C3v | C3v |

`C2z` and both vertical mirrors are valid operations on the 4-atom molecule before the patch; the analyzer simply never asked the principal-axis R2 helper.

## Test plan
- [x] `test_c2v_when_c2_is_not_the_unique_inertia_axis` on the public corrected coordinates at tolerances 0.01, 0.1, 0.3
- [x] Existing `test_cyclic` / `test_symmetric_top_without_rotation_about_the_unique_axis` controls (water C2v, tetramethylhydrazine C2) stay on the same symbols in an isolated overlay of this `analyzer.py`
- [ ] CI `tests/symmetry/test_analyzer.py`


shyuep:

Correct diagnosis and a conservative fix: the fallback only fires when `rot_sym` is still empty, so no existing classification can change, and the test molecule (two axial H, two wings in the *xz* plane) is unambiguously C2v with order 4.

Two things to address:

1. **`_check_R2_axes_asym` searches along inertia-tensor eigenvectors, and this is the symmetric-top path — two of those eigenvectors are degenerate.** Within the degenerate subspace `np.linalg.eigh` returns an arbitrary basis, so whether the C2 lands on a tested axis depends on numerical noise rather than on the molecule. For the specific coordinates in the test it works; a rigidly rotated copy of the same molecule may not be detected. Please add a rotation-invariance test — apply a few fixed random rotations (and small coordinate jitter below tolerance) to the H4 molecule and assert `sch_symbol == "C2v"` each time. If that fails, the fallback needs to search the perpendicular-to-unique-axis plane explicitly instead of relying on the eigenbasis.

2. Related: `_check_R2_axes_asym` is named and documented for the asymmetric-top branch. Calling it from `_proc_sym_top` should carry a one-line comment stating the degeneracy caveat above, otherwise the next reader will assume the axis set is well-defined here.

Otherwise the tolerance sweep (0.01/0.1/0.3) is the right shape for this class of bug, and the CHANGES.md entry is accurate. Tests are green; the `lint` check was still queued at review time.

shyuep (commented):
**Automated PR review generated by Claude** (posted on behalf of @shyuep by a scheduled routine; static analysis + CI only — advisory).

Re-check after `7a0c0a6e`: this commit is a merge of `main` only; the source and test diff are unchanged from the 2026-09-13 review. All five CI jobs (lint + 4-way test matrix) are green post-merge.

The two open items from the previous review still stand: (1) `_check_R2_axes_asym` searches inertia eigenvectors, which are an arbitrary basis in the degenerate subspace of a symmetric top, so a rotation-invariance test (a few fixed random rotations plus sub-tolerance jitter of the H4 molecule, asserting `C2v` each time) is needed to show the fallback is not frame-dependent; (2) a one-line comment at the new call site in `_proc_sym_top` stating that caveat.

PR #147, opened 2026-09-13
Canonicalize symmetry-dataset international symbols

### LINKED ISSUE #139: get_symmetry_dataset(backend="moyopy") disagrees with backend="spglib" on "international" (all 13 monoclinic groups) and "std_origin_shift" (60 groups)

shaneraphel:
## Summary
- Canonicalize the public `international` Hermann–Mauguin value through the existing `SpaceGroup` conversion for both backends.
- Keep `std_origin_shift` and `hall_number` backend-specific; neither is a shared coordinate invariant.
- Assert agreement over space groups 1–230 when the optional `moyopy` dependency is installed.

## Validation
- `python3.12 -m py_compile src/pymatgen/core/structure.py tests/core/test_structure.py`
- Public 230-space-group behavior sweep using pymatgen-core at `25c48ed`, moyopy 0.18.0, and spglib 2.7.0: symbols 13 mismatches → 0; semantic fields 0 mismatches before/after; 60 origin-shift differences deliberately remain.

Closes #139.


shyuep:

The inconsistency in #139 is real, but this fixes it by changing the **established** backend rather than the new one, and that is a silent breaking change to a public field.

1. **`international` changes meaning for every `backend="spglib"` caller.** Per this PR's own (removed) docstring, spglib returns the *short* HM symbol while `SpaceGroup.from_int_number().symbol` returns the *full* one. So `get_symmetry_dataset()["international"]` flips from e.g. `Pnma` to the full HM form for all existing users, none of whom asked for it. Downstream code that string-compares this field or feeds it back into `SpaceGroup(...)` breaks. Preferred direction: canonicalize the **moyopy** path onto spglib's short symbol, so the new backend conforms to the incumbent's convention. If the full symbol is genuinely wanted, add it as a separate key (`international_full`) and leave `international` alone.

2. **The dict becomes internally inconsistent.** `international` is now derived from `dataset.number` alone, discarding any setting/origin-choice information, while `hall_number` in the same dict still carries the backend's actual setting. A consumer that reads both now sees a symbol that does not necessarily correspond to the reported Hall setting. Worth at least a docstring note on `get_symmetry_dataset` that `international` is the canonical symbol for the space-group *type*, not for the setting spglib resolved.

3. **The new assertion cannot fail.** After this change both branches evaluate the identical expression `SpaceGroup.from_int_number(dataset.number).symbol`, so `assert d_moyo["international"] == d_spg["international"]` is a tautology and tests nothing. Add at least one assertion against a hard-coded expected string for a space group where spglib's raw `dataset.international` differed from the canonical symbol — otherwise there is no regression test for the actual canonicalization.

4. Minor: hoisting the `SpaceGroup` import above the branch now runs it on the spglib path too. It is presumably function-local to dodge a circular import, so this is harmless, but it buys nothing either — reverting to the branch-local import keeps the diff to the one line that matters.

CI was still queued at review time; please confirm green before merge.

shyuep:

**Behaviour change worth a second look.** Replacing `dataset.international` with `SpaceGroup.from_int_number(dataset.number).symbol` on the spglib path discards *setting* information. spglib reports the HM symbol for the setting it actually found; `from_int_number` always returns the default setting for that space-group number. For groups with multiple standard settings this silently changes the public output:

- monoclinic unique-axis choices (`P 1 2 1` vs `P 1 1 2`, both #3)
- alternative cell choices (`C 1 2/c 1` vs `I 1 2/a 1`, #15)
- rhombohedral vs hexagonal axes for the R groups

Users who read `["international"]` to know which setting the standardized cell is in now lose that. Since `std_origin_shift` and `hall_number` are deliberately left backend-specific, `international` is arguably the one field that *should* stay setting-aware — the test asserts agreement by construction, so it can no longer catch a real backend disagreement.

Suggestions:
1. If the goal is a backend-invariant label, add it as a new key (e.g. `international_canonical`) and leave `international` as each backend reports it.
2. If the change stands, note it as an API-breaking change and say what replaces the setting-aware symbol.

**Minor**
- The new assert `d_moyo["international"] == d_spg["international"]` is tautological now that both sides compute the same expression from `number`, which is already asserted equal on the line above. It tests nothing. Comparing against `dataset.international` for the spglib path would be the meaningful assertion.
- Moving the `SpaceGroup` import to the top of the branch is fine, but a module-level deferred import would be cleaner if it's now on both paths.

CI: 5/5 checks green.

